PR #3, opened 2021-06-15
Fix integrity issues

yakkomajuri:
Ignores flags that aren't active, and accompanies PostHog/posthog#4764 to fix the issue with the project being able to change.


featureflags.go
```diff
@@ -107,11 +108,17 @@ func (poller *FeatureFlagsPoller) fetchNewFeatureFlags() {
...
-	poller.featureFlags = featureFlagsResponse.Results
+	newFlags := []FeatureFlag{}
+	for _, flag := range featureFlagsResponse.Results {
+		if flag.Active {
+			newFlags = append(newFlags, flag)
+		}
+	}
+	poller.mutex.Lock()
+	poller.featureFlags = newFlags
 	poller.mutex.Unlock()
```
fuziontech, line 122: 👍

fuziontech (approved):

featureflags.go
yakkomajuri: issue wasn't the mutex btw - was just that we weren't filtering for active flags (in any lib 😬 )

fuziontech: no no, I know mutex wasn't issue. This is just more clean locking on just where the mutation was happening.

>(in any lib 😬 )
🥶

yakkomajuri: yeah indeed

PR #5, opened 2022-06-28
feat(libraries): add multivariate support for posthog-go

alexkim205:
Changes:
- [x] Add multivariate support w/ decide v2 endpoint
- [x] Add GetFeatureFlag method
- [x] Add send_feature_flags option to capture

Tests:
- [x] Add a bunch of tests to posthog_test
- [x] Add more e2e tests for feature flags and capture additions


featureflags.go
```diff
@@ -274,3 +255,65 @@ func (poller *FeatureFlagsPoller) ForceReload() {
 func (poller *FeatureFlagsPoller) shutdownPoller() {
 	poller.shutdown <- true
 }
+
+func (poller *FeatureFlagsPoller) getFeatureFlagValue(featureFlag FeatureFlag, key string, distinctId string) (interface{}, error) {
```
liyiy, line 259: `getFeatureFlagValue` is more accurate but since we've already implemented the function as `getFeatureFlag` in our js libraries etc how about keeping it as `getFeatureFlag` for consistency?

liyiy: nvm you did leave it as getFeatureFlag in the client interface it appears, ignore meee 😃

featureflags.go
```diff
@@ -243,6 +219,11 @@ func (poller *FeatureFlagsPoller) GetFeatureFlags() []FeatureFlag {
 func (poller *FeatureFlagsPoller) request(method string, endpoint string, requestData []byte, headers [][2]string) (*http.Response, error) {
 
 	url := poller.Endpoint + "/" + endpoint + "/?token=" + poller.projectApiKey + ""
+
+	if endpoint == "decide" {
```
liyiy, line 223: @neilkakkar @rcmarron do we need the `token` param for the decide v2 endpoint? I've seen it without that param in a server library but I'm not 100% sure

neilkakkar: Good question, you can send it in the POST body too! With the key `api_key`

capture.go
```diff
@@ -69,13 +71,14 @@ func (msg Capture) APIfy() APIMessage {
...
-		Event:          msg.Event,
-		Properties:     myProperties,
+		Type:             msg.Type,
+		Library:          library,
+		LibraryVersion:   libraryVersion,
+		Timestamp:        msg.Timestamp,
+		DistinctId:       msg.DistinctId,
+		Event:            msg.Event,
+		Properties:       myProperties,
+		SendFeatureFlags: msg.SendFeatureFlags,
```
neilkakkar, line 81: I might be getting this completely wrong, but sounds like this is sending a `SendFeatureFlags` property on the event, versus sending the feature properties?

featureflags.go
```diff
@@ -242,8 +220,23 @@ func (poller *FeatureFlagsPoller) GetFeatureFlags() []FeatureFlag {
...
+		poller.Errorf("creating url - %s", err)
+	}
+	searchParams := url.Query()
+
+	if method == "GET" {
+		searchParams.Add("token", poller.projectApiKey)
+	}
+
+	if endpoint == "decide" {
+		searchParams.Add("v", "2")
```
neilkakkar, line 235: this feels too obfuscated, since there's no `decide` anymore, only `decide?v=2`. Thoughts on making either searchParams a parameter, or updating endpoint to be `decide?v=2` . Either should work with `url.Parse()`

posthog.go
```diff
@@ -172,6 +176,25 @@ func (c *client) Enqueue(msg Message) (err error) {
 	case Capture:
 		m.Type = "capture"
 		m.Timestamp = makeTimestamp(m.Timestamp, ts)
+		if m.SendFeatureFlags {
```
neilkakkar, line 179: Ah, I see why!

And this is here because capture doesn't have access to the client. Hmm, this library interface is annoying, with all that enqueuing happening by the client, vs handled internally.

I guess this is fine :)

I'd only ask to remove `sendFeatureFlags` as a key from the message here!

posthog.go
```diff
@@ -452,3 +494,17 @@ func (c *client) notifyFailure(msgs []message, err error) {
...
 }
+
+func (c *client) getFeatureVariants(distinctId string, groups Groups) (map[string]interface{}, error) {
+	if c.featureFlagsPoller == nil {
+		errorMessage := "specifying a PersonalApiKey is required for using feature flags"
+		c.Errorf(errorMessage)
+		return nil, errors.New(errorMessage)
+	}
+
+	requestData, err := c.featureFlagsPoller.getFeatureFlagVariants(distinctId, groups)
```
neilkakkar, line 505:
```suggestion
	featureVariants, err := c.featureFlagsPoller.getFeatureFlagVariants(distinctId, groups)
```

posthog_test.go
```diff
@@ -302,6 +305,7 @@ func TestEnqueue(t *testing.T) {
 					"version":     "1.0.0",
 					"platform":    "macos", // :)
 				},
+				SendFeatureFlags: false,
```
neilkakkar, line 308: Can we have a default `false` so people who don't care about this don't have to insert it?

posthog.go
alexkim205: I can't delete the field directly from the Message object (bc Go says I can't) but I can omit this in the capture API-ify message which should do the same thing!

posthog_test.go
alexkim205: Unfortunately go doesn't support default values :( . However this won't impact the end user because omitting a sendfeatureflags property won't trigger that codepath that injects feature flags.

neilkakkar:
Looks good to me!

PR #6, opened 2022-07-28
Local Feature Flag Evaluation

EDsCODE:
- GetAllFlags
- Adjust IsFeatureEnabled and GetFeatureFlag to share logic
- Local Evaluation supported
- api/feature_flag updated to local_evaluation endpoint
- New suite of tests according to (https://github.com/PostHog/posthog/issues/10459#issuecomment-1193842228)


EDsCODE:
TODO:

- [ ]  default value tests
- [ ] feature_flag_called instrumentation and testing
- [ ]  memory tests

posthog.go
```diff
@@ -37,17 +37,20 @@ type Client interface {
 	Enqueue(Message) error
 	//
 	// Method returns if a feature flag is on for a given user based on their distinct ID
-	IsFeatureEnabled(string, string, bool) (bool, error)
+	IsFeatureEnabled(string, string, bool, Groups, Properties, map[string]Properties) (bool, error)
```
EDsCODE, line 40: We can use [variadic functions](https://gobyexample.com/variadic-functions) here but we'll lose type enforcement

feature_flags_test.go
```diff
@@ -0,0 +1,547 @@
...
+	}))
+	defer server.Close()
+
+	client, _ := NewWithConfig("Csyjlnlun3OzyNJAafdlv", Config{
+		PersonalApiKey: "some very secret key",
+		Endpoint:       server.URL,
+	})
+	defer client.Close()
+
+	results := []bool{false, true, true, false, true}
```
EDsCODE, line 516: This test is super slow. Need to figure out how to have this run properly

featureflags.go
```diff
@@ -142,43 +202,382 @@ func (poller *FeatureFlagsPoller) IsFeatureEnabled(key string, distinctId string
...
+
+	return convertedFirstValue, convertedSecondValue, nil
+
+}
+
+func interfaceToFloat(val interface{}) (float64, error) {
+
+	var i float64
+	switch t := val.(type) {
+	case int:
```
liyiy, line 544: just wondering is there any reason we can't consolidate these case statements since we're doing float64(t) to all of them?

featureflags.go
```diff
@@ -142,43 +202,382 @@ func (poller *FeatureFlagsPoller) IsFeatureEnabled(key string, distinctId string
...
+		valueOrderable, overrideValueOrderable, err := validateOrderable(value, override_value)
+		if err != nil {
+			return false, err
+		}
+
+		return overrideValueOrderable > valueOrderable, nil
+	}
+
+	if operator == "lt" {
+		valueOrderable, overrideValueOrderable, err := validateOrderable(value, override_value)
```
liyiy, line 492: i think we can probably consolidate these two lines higher up

EDsCODE: What do you mean?

feature_flags_test.go
```diff
@@ -0,0 +1,2882 @@
...
+		},
+	)
+
+	if variant != "variant-1" {
+		t.Error("Should match")
+	}
+
+}
+
+func TestCaptureMultipleUsersDoesntOutOfMemory(t *testing.T) {
```
EDsCODE, line 823: I'm unsure how to test this

posthog.go
```diff
@@ -245,21 +259,28 @@ func (c *client) ReloadFeatureFlags() error {
 	return nil
 }
 
-func (c *client) GetFeatureFlag(flagKey string, distinctId string, defaultValue interface{}) (interface{}, error) {
+func (c *client) GetFeatureFlag(flagConfig FeatureFlagPayload) (interface{}, error) {
+	if err := flagConfig.validate(); err != nil {
```
EDsCODE, line 263: Major change is that we accept a featureflag payload and validate it on invocation. This way users don't need to fill in every argument everytime

EDsCODE (commented):
RE: error handling levels. I'm not a go expert but given how it's standard to emit errors back to the top to be handled, I've added a few error logs where it's implicitly handled but otherwise I send the error up the chain so that the featureflag functions will return them

featureflags.go
liyiy: tbh these are just nits but given how overwhelming most of the local feature flags code is the cleaner the code the easier it is to digest but also it's not a high priority since the go library isn't used heavily so 🤷‍♀️

feature_flags_test.go
neilkakkar: Not aware of the intricacies here, but the test is testing that the size of the hash doesn't blow up arbitrarily with new users coming in. (As for long running server-side apps, there's a risk of this running out of memory).

In node, I just check that the distinctIdHasSentFeatureFlagCalls array never exceeds MAX_SIZE, when 100*MAX_SIZE new users are seen:

https://github.com/PostHog/posthog-js-lite/pull/14/files#diff-f8fbefb57e174bbb58200dfeab663d6b65aa93b671b7847f4f48a17bfdf5dd54R272

feature_flags_test.go
```diff
@@ -0,0 +1,2882 @@
...
+		"third-variant",
+		false,
+		false,
+		"third-variant",
+		"first-variant",
+		false,
+		"first-variant",
+	}
+
+	for i := 0; i < 5; i++ {
```
neilkakkar, line 2870:
```suggestion
	for i := 0; i < 1000; i++ {
```

feature_flags_test.go
EDsCODE: I'm testing the sizelimitedmap separately as a unit test. Doesn't seem feasible to test as an integration test since there's no way to expose the map without making it part of the public interface on client which isn't good practice as it's unnecessary to the user

neilkakkar: Works 👍

featureflags.go
```diff
@@ -94,91 +141,454 @@ func (poller *FeatureFlagsPoller) fetchNewFeatureFlags() {
...
+
+		_, exists = groups[groupName]
+
+		if !exists {
+			errMessage := fmt.Sprintf("FEATURE FLAGS] Can't compute group feature flag: %s without group names passed in", flag.Key)
+			return nil, errors.New(errMessage)
+		}
+
+		focusedGroupProperties := groupProperties[groupName]
+		return matchFeatureFlagProperties(flag, distinctId, focusedGroupProperties)
```
neilkakkar, line 289: 🙈 
```suggestion
		return matchFeatureFlagProperties(flag, groups[groupName], focusedGroupProperties)
```

featureflags.go
```diff
@@ -94,91 +141,454 @@ func (poller *FeatureFlagsPoller) fetchNewFeatureFlags() {
...
+	} else {
+		return matchFeatureFlagProperties(flag, distinctId, personProperties)
+	}
+}
+
+func getMatchingVariant(flag FeatureFlag, distinctId string) (interface{}, error) {
+	lookupTable := getVariantLookupTable(flag)
+
+	for _, variant := range lookupTable {
+		minHash, err := _hash(flag.Key, distinctId, "variant")
```
neilkakkar, line 299: minor nit: min and maxHash can be calculated outside the loop, so we need to calc them only once

featureflags.go
neilkakkar: Not sure if this is different in `go`, but if different cases require the same code, you can put them together like so:

```
case int:
case int8:
case int16:
    i = float64(t)
    return i, nil
```

featureflags.go
```diff
@@ -94,91 +141,454 @@ func (poller *FeatureFlagsPoller) fetchNewFeatureFlags() {
...
+		i = float64(t)
+	case uint8:
+		i = float64(t)
+	case uint16:
+		i = float64(t)
+	case uint32:
+		i = float64(t)
+	case uint64:
+		i = float64(t)
+	default:
```
neilkakkar, line 577: just checking, not sure about go syntax: This doesn't always go to `default` right? (it would in other langs, because there's no break/return in the cases above)

posthog.go
```diff
@@ -37,17 +38,20 @@ type Client interface {
 	Enqueue(Message) error
 	//
 	// Method returns if a feature flag is on for a given user based on their distinct ID
-	IsFeatureEnabled(string, string, bool) (bool, error)
+	IsFeatureEnabled(FeatureFlagPayload) (bool, error)
```
neilkakkar, line 41: remember to mention this breaking change in the changelog :)

posthog.go
```diff
@@ -245,21 +260,28 @@ func (c *client) ReloadFeatureFlags() error {
 	return nil
 }
 
-func (c *client) GetFeatureFlag(flagKey string, distinctId string, defaultValue interface{}) (interface{}, error) {
+func (c *client) GetFeatureFlag(flagConfig FeatureFlagPayload) (interface{}, error) {
+	if err := flagConfig.validate(); err != nil {
+		return false, err
+	}
+
 	if c.featureFlagsPoller == nil {
```
neilkakkar, line 268: `personalApiKey` is not required for calling `/decide`, only for local eval

posthog.go
```diff
@@ -245,21 +260,28 @@ func (c *client) ReloadFeatureFlags() error {
...
-		DistinctId: distinctId,
-		Event:      "$feature_flag_called",
-		Properties: NewProperties().
-			Set("$feature_flag", flagKey).
-			Set("$feature_flag_response", flagValue).
-			Set("$feature_flag_errored", err != nil),
-	})
+	flagValue, err := c.featureFlagsPoller.GetFeatureFlag(flagConfig)
+	if *flagConfig.SendFeatureFlagEvents && !c.distinctIdsFeatureFlagsReported.contains(flagConfig.DistinctId, flagConfig.Key) {
+		c.Enqueue(Capture{
```
neilkakkar, line 275: send groups if any : )

posthog.go
```diff
@@ -217,21 +224,29 @@ func (c *client) Enqueue(msg Message) (err error) {
 	return
 }
 
-func (c *client) IsFeatureEnabled(flagKey string, distinctId string, defaultValue bool) (bool, error) {
+func (c *client) IsFeatureEnabled(flagConfig FeatureFlagPayload) (bool, error) {
```
neilkakkar, line 227: this should be bool of `getFeatureFlag`. ?

featureflags.go
```diff
@@ -193,28 +603,37 @@ func (poller *FeatureFlagsPoller) isSimpleFlagEnabled(key string, distinctId str
...
 	if !poller.fetchedFlagsSuccessfullyOnce {
 		<-poller.loaded
 	}
 
-	poller.mutex.RLock()
-
-	defer poller.mutex.RUnlock()
-
 	return poller.featureFlags
 }
```
neilkakkar, line 639: Minor nit: This request method is very confusing (used for both decide and local_evaluation, based on things passed in.

I'd suggest extracting the common bits out, and having two fns, one which calls `/deicde`, and the other for local eval. (they do the auth setup / adding token to URL etc. etc.).

Also note: L691 -> Send group and person properties to decide!

posthog.go
EDsCODE: hmm well here's another consideration to make. the typing is strict here and currently the typing for creating the feature flag object requires a projectapikey. If we want the key to be optional, refactoring needs to happen on how the arguments are accepted

neilkakkar: Fair enough :)

featureflags.go
EDsCODE: nope!

PR #10, opened 2022-08-23
Added ability to manually Flush()

dselans:
This PR extends the client interface to be able to perform a `Flush()` that will
cause the looper to bypass the ticker, queue length checks etc. and send telemetry
data right away.

This is useful when a service shuts down or exits; especially useful for CLI tools
that may have short runtimes.

NOTE: I did not test ctx usage because it would require adding an interface for
the looper (I think?) to cause the looper to wait and that would require a good
chunk of code changes. Another option - maybe to expose a `LoopFunc` to the config
and allow the client to set it but that seems janky because it would exist almost
purely for tests. :|

### DIFF
--- a/config.go
+++ b/config.go
@@ -1,9 +1,10 @@
 package posthog
 
 import (
-	"github.com/xtgo/uuid"
 	"net/http"
 	"time"
+
+	"github.com/xtgo/uuid"
 )
 
 // Instances of this type carry the different configuration options that may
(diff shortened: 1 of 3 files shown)


dselans:
Just realized that I missed that sends occur in a goroutine and that the `Flush` won't block until the send is completed.

Will re-open an updated version.

PR #12, opened 2022-11-11
feat(flags): Add support for variant overrides

neilkakkar:
Following what main branch does here: https://github.com/PostHog/posthog/pull/12652 for local evaluation

### DIFF
--- a/featureflags.go
+++ b/featureflags.go
@@ -14,6 +14,7 @@ import (
 	"strings"
 	"sync"
 	"time"
+	"sort"
 )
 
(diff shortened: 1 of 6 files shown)


featureflags.go
```diff
@@ -573,6 +602,15 @@ func contains(s []interface{}, e interface{}) bool {
 	return false
 }
 
+func containsVariant(variantList []FlagVariant, key string) bool {
```
neilkakkar, line 605: TODO: a way to merge this into the function above?

PR #20, opened 2023-08-01
fix: fixes a nil panic when unable to fetch feature flags

TristanSpeakEasy:
fixes a nil panic when unable to fetch feature flags and `err != nil` which means `res.Body` will be nil at defer time because we aren't returning early enough from the method


TristanSpeakEasy:
apologies for the slight reformatting gofmt had its way with the file let me know if you want me to revert those changes

simplesagar (approved):

neilkakkar:
makes sense, thanks!

PR #22, opened 2023-09-05
fix race condition and some other things

verygoodsoftwarenotvirus:
- fix race condition in `FeatureFlagsPoller`
- remove some useless whitespace, add some useful whitespace
- use errors.As in place of a type assertion
- stop shadowing the `client` struct in tests
- enable race detection in CI test runs

### DIFF
--- a/featureflags.go
+++ b/featureflags.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"net/http"
 	"net/url"
@@ -145,6 +146,7 @@ func (poller *FeatureFlagsPoller) run() {
 func (poller *FeatureFlagsPoller) fetchNewFeatureFlags() {
 	personalApiKey := poller.personalApiKey
 	headers := [][2]string{{"Authorization", "Bearer " + personalApiKey + ""}}
+
 	res, err := poller.localEvaluationFlags(headers)
 	if err != nil || res.StatusCode != http.StatusOK {
 		poller.loaded <- false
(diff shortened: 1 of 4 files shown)


neilkakkar:
Thanks for the contribution @verygoodsoftwarenotvirus :) - could you walk me through what's the race condition please, and why we need another mutex on top of the current one to fix it?

verygoodsoftwarenotvirus:
> could you walk me through what's the race condition please, and why we need another mutex on top of the current one to fix it? 

I tried, and then I realized I couldn't, and in fact I could make use of the pre-existing mutex in a place where it was simply needed, and was able to satisfy the race detector anyway. Thanks.

featureflags.go
```diff
@@ -259,17 +264,16 @@ func (poller *FeatureFlagsPoller) computeFlagLocally(flag FeatureFlag, distinctI
 	}
 
 	if flag.Filters.AggregationGroupTypeIndex != nil {
-
+		poller.mutex.RLock()
```
neilkakkar, line 267: Could you walk me through the error not having this leads to?

And as a follow up, do we need something similar for `poller.GetFeatureFlags()` ?

Otherwise looking good, thanks!

verygoodsoftwarenotvirus: I went to prove the efficacy of my contributions by making all tests run in parallel via `t.Parallel()`, but the truth is this code is highly concurrent and not race-condition free. When I looked through the parts of the code that would need to be guarded by mutexes to make it concurrency safe, it ended up being a significant refactor, and I've decided I have no reason to care about it if ya'll didn't already. :)

neilkakkar: to be clear, I'm not asking you to prove it @verygoodsoftwarenotvirus , just asking you what case you're trying to defend against because I'm unsure.

Worth noting here that a stale read is not a problem, and it's favourable to blocking, since it will eventually update anyway.

What panics is when we read before it being defined, and I don't think that's happening anywhere here (if it does, we need to fix!)

PR #25, opened 2024-01-12
Fix default feature flags polling interval

zaynetro:
Docs specify that default interval is 5 minutes while in practice it is 5 seconds.


neilkakkar:
great catch, thank you!

neilkakkar (approved):

PR #26, opened 2024-01-15
Initialize properties map when sending feature flags

zaynetro:
When capturing an event with SendFeatureFlags set to true and no properties defined posthog panics with nil pointer.

Panic:

```
github.com/posthog/posthog-go.(*client).Enqueue(0x4000abba00, {0x420b130?, 0x40015705a0?) 
  /root/go/pkg/mod/github.com/posthog/posthog-go@v0.0.0-20240110105835-f2ee529330e9/posthog.go:194
```

Example:

```go
client.Enqueue(posthog.Capture{
	DistinctId: user.ID,
	Event:      "test_event",
	SendFeatureFlags: true,
})
```

### DIFF
--- a/posthog.go
+++ b/posthog.go
@@ -189,6 +189,11 @@ func (c *client) Enqueue(msg Message) (err error) {
 			if err != nil {
 				c.Errorf("unable to get feature variants - %s", err)
 			}
+
+			if m.Properties == nil {
+				m.Properties = NewProperties()
+			}
+
 			for feature, variant := range featureVariants {
 				propKey := fmt.Sprintf("$feature/%s", feature)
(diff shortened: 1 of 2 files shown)


posthog.go
```diff
@@ -189,6 +189,11 @@ func (c *client) Enqueue(msg Message) (err error) {
 			if err != nil {
 				c.Errorf("unable to get feature variants - %s", err)
 			}
+
+			if m.Properties == nil {
```
neilkakkar, line 193: great catch, thanks @zaynetro , could you please also add a testcase for sendFeatureFlags in capture?

neilkakkar:
lovely, thanks!

PR #31, opened 2024-02-05
feat(config): disable geo ip and it is true by default, because on no…

o1o1o3:
Add geo disable parameter, because our users get IP from server and we want to control this behaviour.
This config we made default is true, because same behaviour in nodejs library (https://github.com/PostHog/posthog-js-lite/commit/84df81318a4dd11e481ff72119a38e7c3695b130).

Please review and lets merge it!

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -8,10 +8,10 @@ jobs:
     name: Build
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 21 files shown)


neilkakkar:
Thanks @dikkini , sorry this got overlooked, will take a look next week!

JoshGlazebrook:
Any update on this?

sqadan:
What's the update on this?

PR #41, opened 2024-07-20
Chore: Correct the questions link

jasperchess:
The community Slack has been deprecated. This PR points users to (what I think is) the next best option


ivanagas (approved):
Good catch, thanks!

PR #47, opened 2024-08-08
fix: change version from 2.0 to 1.2 (we are not on v2 yet)

fuziontech:
Golang requires a `v2` folder for major versions like `2.0.0`. We don't really want to do that yet since we are not any major versions ahead of 1.0 yet and have never really used SEMVER.

Let's just reset this to `v1.2.1`


dmarticus (approved):
never stamped more blindly in my life

PR #48, opened 2024-08-08
Better error for feature flags fetch without errors. Adds missing format string verbs to poller Errorf.

mkraft:
We get error responses from the API with no details about why (e.g. rate limiting) and we'd like to see that HTTP response payload appear in the error log rather than what's currently displayed:

```
ERROR: Unable to fetch feature flags%!(EXTRA <nil>)
```

Also, there were several invocations of `poller.Errorf` that were missing the Go format string verb (`%s`).


dmarticus (approved):
Looks great, thanks for your contribution!

PR #50, opened 2024-08-08
feat: allow labels to bump patch/minor versions

fuziontech:
We need to make sure that every land to master bumps the version so people can pin a version with confidence. This allows that.

### DIFF
--- a/.github/workflows/bump-version.yml
+++ b/.github/workflows/bump-version.yml
@@ -1,0 +1,59 @@
+name: Bump Version
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  bump-version:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Set up Go
+        uses: actions/setup-go@v3
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/bump-version.yml
```diff
@@ -0,0 +1,59 @@
...
+          go-version: '1.x'
+
+      - name: Determine version bump type
+        id: bump-type
+        run: |
+          if [[ $(GITHUB_TOKEN=${{ secrets.POSTHOG_BOT_GITHUB_TOKEN }} gh pr view ${{ github.event.pull_request.number }} --json labels -q '.labels[].name' | grep -q "minor") ]]; then
+            echo "type=minor" >> $GITHUB_OUTPUT
+          else
+            echo "type=patch" >> $GITHUB_OUTPUT
+          fi
```
dmarticus, line 28: good call not supporting major version bumps; I don't want to deal with v2 yet.

dmarticus (approved):
yee

.github/workflows/bump-version.yml
fuziontech: 1000% - that will be a manual process anyways

PR #51, opened 2024-08-08
fix: Use the default github action token for pushing

fuziontech:
- Use the default GITHUB_TOKEN for gh cli
- Use a new Personal Access Token for getting around branch rules requiring a PR


dmarticus (approved):
good catch

PR #52, opened 2024-08-08
fix: Use github token to checkout - hopefully setting the right permissions

fuziontech:
try try try again

basing this attempt off of:
https://github.com/orgs/community/discussions/25305


dmarticus (approved):
image

PR #65, opened 2024-08-14
Ability to specify uuid when capturing events

evgeek:
Hello!

We really miss the ability to specify the uuid of events manually when capturing events.
The thing is that we cannot guarantee that events come to our system only once. In the current implementation, in such cases, events in Posthog will be duplicated.
In the SDK in other languages, we got around this by manually generating the uuid, calculated based on the immutable fields of the event - however, in the Golang SDK, unfortunately, there is no such option yet.
Please consider the PR adding this functionality.

P.S. As far as I can see, the current implementation does not use the `posthog.Config.uid()` function, so I removed it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 1.2.18
+
+* [Full Changelog](https://github.com/PostHog/posthog-go/compare/v1.2.17...v1.2.18)
+
 ## 1.2.17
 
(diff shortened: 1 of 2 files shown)


migueldv90:
I think this pull request broke posthog.Capture. I am currently getting an error in v1.2.19.
posthog 2024/08/24 23:59:13 INFO: response 400 400 Bad Request – failed to parse request: data did not match any variant of untagged enum RawRequest

I had to downgrade to v1.2.18 and everything is working as expected.

Issue #69, opened 2024-08-25
v1.2.19 broke posthog.Capture


migueldv90:
I am currently integrating Posthog in a go environment. It seems that from version v1.2.18 to v1.2.19 posthog.Capture broke. I was able to use posthog.Identify without problem but was not able to use posthog.Capture. I had to downgrade to v1.2.18 and everything seems to work as expected.

The error I am getting is -
posthog 2024/08/24 23:59:13 INFO: response 400 400 Bad Request – failed to parse request: data did not match any variant of untagged enum RawRequest

I was not able to pinpoint the specific problem. But I think it has to do with "Ability to specify uuid when capturing events" #68

tylergannon:
Yeah we auto-updated to 1.2.19 and it took down our event monitoring.  Had to roll back to 1.2.18.

migueldv90:
It looks like they have fixed it. I will confirm later and will close this issue when I can verify.

dmarticus:
@migueldv90 can you confirm that this has been fixed for you?  Just doing some housekeeping on our issues and I think this has been resolved, but wanted to double-check.

dmarticus:
Closing since I tested this myself, but feel free to reopen if I'm wrong

migueldv90:
@dmarticus 
Yes, this is working.

Thank you!

PR #70, opened 2024-08-25
Hotfix capture with empty UUID

evgeek:
Hotfix for https://github.com/PostHog/posthog-go/issues/69


posthog.go
```diff
@@ -462,7 +462,7 @@ func (c *client) send(msgs []message) {
 
 // Upload serialized batch message.
 func (c *client) upload(b []byte) error {
-	url := c.Endpoint + "/batch/"
+	url := c.Endpoint + "/capture/"
```
dmarticus, line 465: wait, I don't understand; you're removing the `/batch` functionality?  I don't think that's a good long-term fix for this; folks are using that piece, too.  If we're looking to fix something quickly, we should just revert the change and then fix it more permanently instead of continuing to change more things.

evgeek: Yes, it is just a quick (not good) fix, i describe my thoughts here - https://github.com/PostHog/posthog-go/pull/68#issuecomment-2308798201.

Do you think you can configure `/batch` on the server side to work with an empty uuid, like `/capture`? If (I hope) this works out, previous PR will be ok. If not, we will need to think about how best to implement the functionality from the SDK side.

But for now you're right - rollback is better. I don't know all the SDK use cases and can't be sure it won't break something else.
Rollback PR - https://github.com/PostHog/posthog-go/pull/71

PR #71, opened 2024-08-26
Rollback capture with empty UUID

evgeek:


dmarticus (approved):
Thank you!

evgeek:
With pleasure :)

@dmarticus, do you think I can expect a fix for `/batch` behavior with empty uuid in the near future? Or is it not your responsibility and I should propose this fix elsewhere? (where? :))

Or maybe it would be better to just implement this feature at the SDK level in some other way?

PR #84, opened 2025-02-11
feat(flags): Add method for fetching decrypted remote config flag payload

havenbarnes:
Context: [PostHog/posthog#27414 (files)](https://github.com/PostHog/posthog/pull/27414/files)


posthog.go
```diff
@@ -679,6 +686,44 @@ func (c *client) makeDecideRequest(distinctId string, groups Groups, personPrope
...
+	}
+
+	req, err := http.NewRequest("GET", url.String(), nil)
+	if err != nil {
+		return "", fmt.Errorf("creating request: %v", err)
+	}
+
+	req.Header.Set("Authorization", "Bearer "+c.PersonalApiKey)
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("User-Agent", "posthog-go (version: "+Version+")")
```
marandaneto, line 703: I'd keep it as `${sdk}/${version}` eg 'posthog-go/1.0.0' so its easier to parse the user agent whenever we need to do something specific per SDK

havenbarnes: Ok cool, I'll change this everywhere in the SDK, I just copied it from the decide request

haacked: For `posthog-dotnet` I've followed the format `sdk/version (extra info)`. More specifically: `$"{LibraryName}/{VersionConstants.Version} ({framework}; {os}; {arch})"`

dmarticus (approved):

featureflags.go
```diff
@@ -900,7 +900,7 @@ func (poller *FeatureFlagsPoller) request(method string, url *url.URL, requestDa
 
 	version := getVersion()
 
-	req.Header.Add("User-Agent", "posthog-go (version: "+version+")")
+	req.Header.Add("User-Agent", "posthog-go/"+version)
```
marandaneto, line 903: nitpick but `posthog-go` can be a const since its being used many times over

marandaneto (approved):
I don't know what "decrypted remote config flag payload" is but codewise looks good

feature_flags_test.go
```diff
@@ -945,6 +945,33 @@ func TestGetFeatureFlagPayload(t *testing.T) {
...
+
+	payload, _ := client.GetDecryptedFeatureFlagPayload(1)
+
+	var payloadMap map[string]interface{}
+	err := json.Unmarshal([]byte(payload), &payloadMap)
+	if err != nil {
+		t.Error("Failed to decode payload")
+	}
+
+	if payloadMap["foo"] != "bar" || payloadMap["baz"] != float64(42) {
```
haacked, line 969: I assume the range of types that the values can be are limited to `string|number|boolean`, correct?

haacked: Ok, scratch that. As I understand it, the payload can be arbitrary JSON. Might be good to have an example that shows how to respond to a JSON array payload.

featureflags.go
havenbarnes: Done!

PR #85, opened 2025-02-21
feat(feature-flags): support quota limiting for feature flags

dmarticus:
with https://github.com/PostHog/posthog/pull/28564, we now start to respond different with the /decide and /local_evaluation APIs if users have gone over their quota limit. Now we need to change the SDKs to handle these new responses.


posthog.go
```diff
@@ -730,6 +730,15 @@ func (c *client) getFeatureFlagFromDecide(key string, distinctId string, groups
...
 
+	// Check if feature_flags is quota limited
+	if decideResponse.QuotaLimited != nil {
+		for _, limitedFeature := range *decideResponse.QuotaLimited {
+			if limitedFeature == "feature_flags" {
+				return false, nil
+			}
+		}
+	}
+
```
haacked, line 741: Looks like this code is duplicated in a couple places. Perhaps a helper method `isQuotaLimited(decideResponse)` (or something like that) would make sense here?

haacked (approved):
Based on my limited Go understanding, this looks good to me. One minor suggestion to refactor common code into a method, but not urgent.

PR #86, opened 2025-03-26
feat: increase message size limit to 0.5Mb

k11kirky:
The max allowed size of our events is 1Mb so we should not limit the size to 32kb in the SDK.


message.go
```diff
@@ -120,5 +120,5 @@ func (q *messageQueue) flush() (msgs []message) {
 
 const (
 	maxBatchBytes   = 500000
-	maxMessageBytes = 32000
+	maxMessageBytes = 500000
```
orian, line 123: why not 1MB?

orian (approved):

PR #89, opened 2025-03-30
refactor: Consolidate decide code

haacked:
There were two implementations of decide. That's not the end of the world, but I'm about to implement a new decide format (`/decide?v=4`) and I don't want to have to do it in two places.

So this adds a new file `decide.go` with a new interface `decider` and an implementation `decideClient` that both `Client` and `FeatureFlagsPoller` uses.

### DIFF
--- a/decide.go
+++ b/decide.go
@@ -11,6 +11,20 @@ import (
 	"time"
 )
 
+type DecideRequestData struct {
+	ApiKey           string                `json:"api_key"`
+	DistinctId       string                `json:"distinct_id"`
+	Groups           Groups                `json:"groups"`
+	PersonProperties Properties            `json:"person_properties"`
+	GroupProperties  map[string]Properties `json:"group_properties"`
+}
+
+type DecideResponse struct {
+	FeatureFlags        map[string]interface{} `json:"featureFlags"`
+	FeatureFlagPayloads map[string]string      `json:"featureFlagPayloads"`
+	QuotaLimited        *[]string              `json:"quota_limited"`
+}
+
 // decider defines the interface for making decide requests
 type decider interface {
 	makeDecideRequest(distinctId string, groups Groups, personProperties Properties, groupProperties map[string]Properties) (*DecideResponse, error)
--- a/decide.go
+++ b/decide.go
@@ -1,0 +1,96 @@
+package posthog
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"time"
+)
+
+// decider defines the interface for making decide requests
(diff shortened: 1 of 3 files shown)


decide.go
```diff
@@ -0,0 +1,96 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("sending request: %v", err)
+	}
+	defer res.Body.Close()
+
+	if res.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("unexpected status code from /decide/: %d", res.StatusCode)
+	}
+
+	resBody, err := ioutil.ReadAll(res.Body)
```
haacked, line 84: `ioutil.ReadAll` is deprecated: As of Go 1.16, this function simply calls `[io.ReadAll]`. I'll make this change.

haacked:
I need to make a couple changes to this PR. I'll move it to DRAFT.

decide.go
```diff
@@ -0,0 +1,110 @@
...
+	Groups           Groups                `json:"groups"`
+	PersonProperties Properties            `json:"person_properties"`
+	GroupProperties  map[string]Properties `json:"group_properties"`
+}
+
+type DecideResponse struct {
+	FeatureFlags        map[string]interface{} `json:"featureFlags"`
+	FeatureFlagPayloads map[string]string      `json:"featureFlagPayloads"`
+	QuotaLimited        *[]string              `json:"quota_limited"`
+}
```
dmarticus, line 26: shouldn't this have `errorsWhileComputingFlags`?  How are you modeling that?

decide.go
```diff
@@ -0,0 +1,110 @@
...
+		return nil, fmt.Errorf("error reading response from /decide/: %v", err)
+	}
+
+	var decideResponse DecideResponse
+	err = json.Unmarshal(resBody, &decideResponse)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing response from /decide/: %v", err)
+	}
+
+	return &decideResponse, nil
```
dmarticus, line 109: you still need to handle the most common error case – decide returns `errorsWhileComputingFlags`, but still returns a 200.

dmarticus: I see that you're pulling from an existing implementation that doesn't handle that response either, so, I get why you didn't do it. I just think it's worth doing.

dmarticus (commented):
I think this needs to handle the case where `/decide` returns `errorsWhileComputingFlags`, seems like an oversight in the Go SDK that we didn't handle this at all.

dmarticus:
Ah, I see that you do this handling for that new error field in https://github.com/PostHog/posthog-go/pull/91, so I'm fine with you just doing it there.  Stamping.

dmarticus (approved):

PR #91, opened 2025-04-01
chore(flags): Decide v4 support - adds version, id, reason, and requestId to `$feature_flag_called` events

haacked:
This is follow on work for https://github.com/PostHog/posthog/issues/29693. The work here is similar to https://github.com/PostHog/posthog-python/pull/207.

## Updates to the `$feature_flag_called` event

When capturing the `$feature_flag_called` event, additional information are now captured:

The end result is `$feature_flag_called` events will have additional properties:

| Property Name                |  Descirption
| ------------------------ | -------------
| `$feature_flag_version` | The version of the feature flag. This is visible in the diffs of the feature flag activity log.
| `$feature_flag_reason` | The reason the feature flag matched or didn't match.
| `$feature_flag_id`          | The database id of the feature flag (Just in case a flag was deleted and then recreated with the same key, we can determine the difference.

## Backwards compatibility:

The changes are all backwards compatible with `/decide?v=3`.

## Testing

Unit tests and some manual testing


decide.go
```diff
@@ -19,10 +19,137 @@ type DecideRequestData struct {
...
+
+// CommonResponseFields contains fields common to all decide response versions
+type CommonResponseFields struct {
+	QuotaLimited              *[]string `json:"quota_limited"`
+	RequestId                 string    `json:"requestId"`
+	ErrorsWhileComputingFlags bool      `json:"errorsWhileComputingFlags"`
+}
+
+// UnmarshalJSON implements custom unmarshaling to handle both v3 and v4 formats
+func (r *DecideResponse) UnmarshalJSON(data []byte) error {
```
dmarticus, line 101: this is a great abstraction, very idiomatic golang

dmarticus (approved):
looks great!

decide.go
haacked: Thanks!

PR #97, opened 2025-05-03
Simplify executor implementation

eli-r-ph:
Just a small refactor to make the Go task execution faster/cheaper without changing the core behavior.

### DIFF
--- a/executor.go
+++ b/executor.go
@@ -13,7 +13,7 @@ func newExecutor(cap int) *executor {
 	return e
 }
 
-func (e *executor) do(task func()) (ok bool) {
+func (e *executor) do(task func()) bool {
 	select {
 	case e.queue <- task:
 		// task is enqueued successfully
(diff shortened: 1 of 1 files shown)


executor.go
```diff
@@ -37,17 +32,7 @@ func (e *executor) close() {
 
 func (e *executor) loop() {
 	for task := range e.queue {
-		go e.run(task)
+		capturedTask := task
+		go capturedTask()
 	}
```
eli-r-ph, line 37: ☝️ until Go 1.24 (this library is on 1.18 atm) the loop variable used in this construct is reused per iteration - this just prevents a race when passing the tasks along to goroutines when more than one can come off the queue at once, as will happen occasionally with the change to a buffered channel.

mariusandra (approved):
The changes look legit, though I lack the context to fully evaluate them. The only time I used go was when migrating this library 5 years ago... so I trust you know what you're doing 👍

benjackwhite (approved):
Low context 👍

Issue #80, opened 2024-12-16
Fatal error on GetFeatureFlag: concurrent map writes


albertteoh:
Our application crashed with the following error:

```
fatal error: concurrent map writes

goroutine 7975 [running]:
github.com/posthog/posthog-go.(*SizeLimitedMap).add(...)
	/src/vendor/github.com/posthog/posthog-go/util.go:26
github.com/posthog/posthog-go.(*client).GetFeatureFlag(0xc000b7a100, {{0x1e0b64c, 0x2d}, {0x1dbe3a5, 0xc}, 0xc0017f8780, 0xc0017f87b0, 0xc0017f87e0, 0x0, 0xc0005abfb0})
```

It appears there was a concurrent write on the `sizeLimitedMap.ids` map. 

Go's maps do not support safe concurrent writes.

jsirianni:
I am seeing this in my application as well.

PR #99, opened 2025-05-13
chore: remove own cache implementation, use a thread safe LRU cache

orian:
Fixes the issue with concurrent access to getFeatureFlag race conditions.

Using the LRU with key: distinctID, flagKey will limit the max size for real.

Also, the cache won't be all cleared when reaching max size, so no surges in requests.|

Fixes: #80


posthog.go
```diff
@@ -370,13 +382,14 @@ func (c *client) GetFeatureFlag(flagConfig FeatureFlagPayload) (interface{}, err
 			}
 		}
 
-		c.Enqueue(Capture{
+		if c.Enqueue(Capture{
```
rafaeelaudibert, line 385: Do we need this extra if?

PR #102, opened 2025-05-20
fix: the local evaluation was not using rollout percentage

orian:


featureflags_test.go
```diff
@@ -0,0 +1,26 @@
...
+func TestCalculateHash(t *testing.T) {
+	for _, tt := range []struct {
+		ident string
+		want  float64
+	}{
+		{"some_distinct_id", 0.086639},
+		{"test-identifier", 0.749634},
+		{"example_id", 0.869139},
+		{"example_id2", 0.844273},
+	} {
```
haacked, line 17: These look like the same cases we use in https://github.com/PostHog/posthog/blob/master/rust/feature-flags/src/flags/flag_matching_utils.rs#L556-L570 but the expected values are different. These should have the same expected values.

featureflags.go
```diff
@@ -831,27 +822,16 @@ func containsVariant(variantList []FlagVariant, key string) bool {
...
 // extracted as a regular func for testing purposes
-func checkIfSimpleFlagEnabled(key string, distinctId string, rolloutPercentage uint8) (bool, error) {
-	val, err := _hash(key, distinctId, "")
-	if err != nil {
-		return false, err
-	}
-
-	return val <= float64(rolloutPercentage)/100, nil
+func checkIfSimpleFlagEnabled(key, distinctId string, rolloutPercentage uint8) bool {
+	hash := calculateHash(key, distinctId, "")
```
haacked, line 826: Per the below comment, I believe this should be:

```suggestion
	hash := calculateHash(key + ".", distinctId, "")
```

If you make the suggested change below.

featureflags.go
```diff
@@ -831,27 +822,16 @@ func containsVariant(variantList []FlagVariant, key string) bool {
...
+	return hash <= float64(rolloutPercentage)/100
 }
 
-func _hash(key string, distinctId string, salt string) (float64, error) {
+func calculateHash(key, distinctId, salt string) float64 {
 	hash := sha1.New()
-	hash.Write([]byte("" + key + "." + distinctId + "" + salt))
+	hash.Write([]byte(key + "." + distinctId + salt))
```
haacked, line 832: Looking at https://github.com/PostHog/posthog/blob/master/rust/feature-flags/src/flags/flag_matching_utils.rs#L41-L49, I believe the "." here is incorrect. Notice the method there does this:

```rust
let hash_key = format!("{}{}{}", prefix, hashed_identifier, salt);
```

What your method calls `key`, the Rust implementation calls `prefix`. If a `.` is needed, then the `.` should be part of the prefix.

You can see that here: https://github.com/PostHog/posthog/blob/master/rust/feature-flags/src/flags/flag_matching.rs#L1149

```rust
calculate_hash(&format!("{}.", feature_flag.key), &hashed_identifier, salt)
```

```suggestion
	hash.Write([]byte(key + distinctId + salt))
```

But this change requires the above corresponding change.

haacked (changes requested):

haacked:
I think a few minor fixes will make this ready for prime time! Thanks for this!

featureflags.go
orian: The go version is IMO cleaner and less error prone, joins strings just once ;)
In rust one has to make sure that prefix is correct.
I'm gonna switch to Rust version, as i expect us to migrate holdouts soon.

PR #107, opened 2025-05-28
fix: race in accessing feature flags

orian:


featureflags.go
```diff
@@ -852,10 +854,12 @@ func calculateHash(prefix, distinctId, salt string) float64 {
 
 func (poller *FeatureFlagsPoller) GetFeatureFlags() ([]FeatureFlag, error) {
 	// When channel is open this will block. When channel is closed it will immediately exit.
-	_, closed := <-poller.loaded
```
orian, line 855: closed was always false at this place

featureflags.go
```diff
@@ -852,10 +854,12 @@ func calculateHash(prefix, distinctId, salt string) float64 {
...
-	if closed && poller.featureFlags == nil {
+	<-poller.firstFeatureFlagRequestFinished
+	poller.mutex.RLock()
```
orian, line 858: here was a data race

haacked (approved):
:shipit: looks like a typo in a comment but otherwise looks good.

PR #109, opened 2025-06-10
chore: move URL parsing errors to init

orian:
* move feature flag URL parsing to initialization,
* proper logger in flags (allow to lower the message level to warn)


executor.go
```diff
@@ -1,38 +1,43 @@
...
 func (e *executor) do(task func()) bool {
-	select {
-	case e.queue <- task:
-		// task is enqueued successfully
-		return true
+	e.m.Lock()
+	defer e.m.Unlock()
 
-	default:
-		// buffer was full; inform the caller rather than blocking
```
orian, line 23: This buffer was never full, because in line 36 we were starting a new goroutine for each added task. The error here could happen only if the rate of starting goroutines would be lower than adding new tasks and I think this never happen to us.

Also, the test was incorrect.

config.go
```diff
@@ -123,9 +124,9 @@ const (
 	DefaultBatchSize = 250
 )
 
-// Verifies that fields that don't have zero-values are set to valid values,
+// Validate verifies that fields that don't have zero-values are set to valid values,
```
dmarticus, line 127:
```suggestion
// Validates that fields that don't have zero-values are set to valid values,
```

dmarticus (approved):
nice work!

config.go
dmarticus: oh wow TIL.  Thank you!

PR #118, opened 2025-08-07
fix(flags): Pass project API key in remote_config requests

haacked:
Fixes the remote_config endpoint to include the project API key as a token parameter for deterministic project routing, matching the implementation in other server-side SDKs.

- Add token parameter to remote_config URL construction
- Add test to verify the token parameter is included in requests

See https://github.com/PostHog/posthog/issues/35303 for the problem.
Port of PostHog Python implementation: https://github.com/PostHog/posthog-python/pull/303

### DIFF
--- a/posthog.go
+++ b/posthog.go
@@ -711,7 +711,7 @@ func (c *client) getFeatureVariantsWithOptions(distinctId string, groups Groups,
 }
 
 func (c *client) makeRemoteConfigRequest(flagKey string) (string, error) {
-	remoteConfigEndpoint := fmt.Sprintf("api/projects/@current/feature_flags/%s/remote_config/", flagKey)
+	remoteConfigEndpoint := fmt.Sprintf("api/projects/@current/feature_flags/%s/remote_config?token=%s", flagKey, c.key)
 	url, err := url.Parse(c.Endpoint + "/" + remoteConfigEndpoint)
 	if err != nil {
 		return "", fmt.Errorf("creating url: %v", err)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 1.6.2
+
+* Fix: Pass project API key in remote_config requests
+* [Full Changelog](https://github.com/PostHog/posthog-go/compare/v1.6.1...v1.6.2)
+
 ## 1.6.1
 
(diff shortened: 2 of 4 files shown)


posthog.go
```diff
@@ -711,7 +711,7 @@ func (c *client) getFeatureVariantsWithOptions(distinctId string, groups Groups,
 }
 
 func (c *client) makeRemoteConfigRequest(flagKey string) (string, error) {
-	remoteConfigEndpoint := fmt.Sprintf("api/projects/@current/feature_flags/%s/remote_config/", flagKey)
+	remoteConfigEndpoint := fmt.Sprintf("api/projects/@current/feature_flags/%s/remote_config?token=%s", flagKey, c.key)
```
orian, line 714: You should set the key on the parsed url, it has a dictionary for it,

posthog.go
```diff
@@ -711,13 +711,21 @@ func (c *client) getFeatureVariantsWithOptions(distinctId string, groups Groups,
 }
 
 func (c *client) makeRemoteConfigRequest(flagKey string) (string, error) {
-	remoteConfigEndpoint := fmt.Sprintf("api/projects/@current/feature_flags/%s/remote_config/", flagKey)
-	url, err := url.Parse(c.Endpoint + "/" + remoteConfigEndpoint)
+	baseURL, err := url.JoinPath(c.Endpoint, "api/projects/@current/feature_flags", flagKey, "remote_config")
```
orian, line 714: url.JoinPath is probably available only in some new versions of the std library, 

In general, posthog-go is tested against some ancient versions of Go, I've checked and most companies (including Go) support ~2 previous versions.

The Go states:
>Each major Go release is supported until there are two newer major releases. For example, Go 1.5 was supported until the Go 1.7 release, and Go 1.6 was supported until the Go 1.8 release. We fix critical problems, including [critical security problems](https://go.dev/security), in supported releases as needed by issuing minor revisions (for example, Go 1.6.1, Go 1.6.2, and so on).

and I think we should do something similar (or maybe 1/2 more version or so)

orian (approved):

posthog.go
haacked: Although I'm reading that 1.21 has enhanced back compat features. I think I'll go with your recommendation @orian and move us to `1.21`.

orian: <3 sounds great to me

PR #120, opened 2025-08-12
feat: Introduce error tracking with slog integration

vdekrijger:
## Summary

The code in this PR introduces [error tracking](https://posthog.com/docs/error-tracking/installation/manual) within the Go SDK. This removes the need to rely on manual API calls to propagate errors towards the PostHog system. Additionally it also introduces a [slog](https://go.dev/blog/slog) wrapper, increasing the convenience for using this within Go applications.

## Context / Motivation

The reason why I added this change, is that I was looking into integrating PostHog for my side project, where I have a microservice architecture with backend services written in both Go and Python. PostHog is looking promising for my Python / React services, however for Go there was no convenient way to capture errors. Wanting to avoid using several different services for different components, I opted to introduce this to the library with the hopes of helping someone else out as well 😃. 

For the slog wrapper, I’ve opted to use an instantiated logger rather than modifying the global state. This approach gives developers more control over their dependencies, avoids implying that global state is preferred, and encourages a safer default usage pattern.

## Proof of work

The code can be used as following (either one of the 2 ways work): 

Scherm­afbeelding 2025-08-12 om 22 51 23

_Code obtained from [here](https://github.com/vassiedk/posthog-go/blob/369a008f0e21d80d3b28dd18a1b1e6af727e8ba5/README.md?plain=1#L71-L87)_

- ✅ All tests (including the new ones) are passing.
- Running the example code looks as following:
    - <details>
       <summary>1. Running the example/main.go code (other functions commented out to avoid clutter)</summary>

       Scherm­afbeelding 2025-08-12 om 22 05 15

        </details>
    - <details>
       <summary>2. Logs from `TestErrorTrackingThroughEnqueueing`</summary>

       Scherm­afbeelding 2025-08-12 om 22 05 49

        </details>
    - <details>
       <summary>3. Logs from `TestErrorTrackingThroughLogHandler`</summary>

       Scherm­afbeelding 2025-08-12 om 22 06 02

        </details>

    - <details>
       <summary>4. Overview of the errors within the UI</summary>

        Scherm­afbeelding 2025-08-12 om 22 06 26

        </details>
    - <details>
       <summary>5. Overview of the detail page of an error</summary>

        Scherm­afbeelding 2025-08-12 om 22 06 56

        </details>
    - <details>
       <summary>6. All stack frames rendering correctly and respecting the `inApp` flag</summary>

        Scherm­afbeelding 2025-08-12 om 22 07 15
        Scherm­afbeelding 2025-08-12 om 22 07 30
[...]

PR #122, opened 2025-08-26
chore: Bump Go version in CI to 1.21 to align with `go.mod`

vassiedk:
## Context

This PR bumps the Golang version in the CI to [version 1.21](https://tip.golang.org/doc/go1.21). This way the version is aligned with the version defined in the `go.mod` file ([code](https://github.com/PostHog/posthog-go/blob/master/go.mod)). Additionally it will unblock [this PR](https://github.com/PostHog/posthog-go/pull/120) as it depends on features introduced in `Go 1.21`.

Both version 1.19 and version 1.21 are deprecated:

Scherm­afbeelding 2025-08-26 om 18 00 53

_([source](https://endoflife.date/go))_

However I figured that bumping the version to the lowest required version (albeit EOL) to unblock other work, will keep the library usable for the widest possible audience.

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -9,10 +9,10 @@ jobs:
     name: Build
     runs-on: ubuntu-latest
     steps:
-      - name: Set up Go 1.19
+      - name: Set up Go 1.21
         uses: actions/setup-go@v1
         with:
-          go-version: 1.19
+          go-version: 1.21
 
       - name: Check out source code
         uses: actions/checkout@v1
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
 # PostHog Go
 
 [![Go Reference](https://pkg.go.dev/badge/github.com/posthog/posthog-go.svg)](https://pkg.go.dev/github.com/posthog/posthog-go)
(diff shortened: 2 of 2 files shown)


daibhin (approved):
Looks good. Official source for version deprecations can be found at: https://go.dev/doc/devel/release#policy

vassiedk:
I was looking into the [failed CI job](https://github.com/PostHog/posthog-go/actions/runs/17243728135/job/48976360899?pr=122), and it looks like the [workflow code](https://github.com/PostHog/.github/blob/d3a6ddf832c782be83e56bfe685c13f69a1cf473/.github/workflows/flags-project-board.yml#L74) depends upon the `github-token: ${{ secrets.POSTHOG_PROJECT_BOT_TOKEN }}`.

Since this PR is coming from a fork, that secret isn’t available (as expected), which causes the actions/github-script step to fail with:

> Error: Input required and not supplied: github-token

From what I’ve gathered (through a [GitHub discussion](https://github.com/orgs/community/discussions/50161), [a blog post](https://michaelheap.com/access-secrets-from-forks/) and [a stackoverflow post](https://stackoverflow.com/questions/76975408/how-to-pass-a-secret-from-a-forked-github-repo-to-source-repo)), this is a common limitation when forks trigger workflows that rely on repo/org secrets.

@daibhin Do you have a standard process for handling this scenario on PRs from forks? Is there something I can do to help unblock this?

PR #131, opened 2025-11-07
feat: Add SDK parity for date operators and inconclusive evaluation

dustinbyrne:
- Fixed bug where inconclusive feature flag evaluations returned `false` instead of `nil`: When local evaluation couldn't determine a flag's value due to missing properties, it returned `false` (with an error), potentially causing flags to appear disabled when they should be inconclusive. Now properly returns `nil` with an error.
- Added date comparison support for local feature flag evaluation: Implemented `is_date_before` and `is_date_after` operators for property matching, supporting both absolute dates (most common ISO 8601 timestamps) and relative dates (e.g., -7d, 2w, 3m, 1y)

### DIFF
--- a/featureflags.go
+++ b/featureflags.go
@@ -20,6 +20,8 @@ import (
 
 const LONG_SCALE = 0xfffffffffffffff
 
+var relativeDateRegex = regexp.MustCompile(`^-?([0-9]+)([hdwmy])$`)
+
 type FeatureFlagsPoller struct {
 	// firstFeatureFlagRequestFinished is used to log feature flag usage before the first feature flag request is done.
 	// After the request the channel get closed.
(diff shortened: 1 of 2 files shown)


featureflags.go
```diff
@@ -638,7 +640,7 @@ func (poller *FeatureFlagsPoller) matchFeatureFlagProperties(
 	}
 
 	if isInconclusive {
-		return false, &InconclusiveMatchError{"Can't determine if feature flag is enabled or not with given properties"}
+		return nil, &InconclusiveMatchError{"Can't determine if feature flag is enabled or not with given properties"}
```
dmarticus: why did this return type change?

dmarticus (approved):
one question on a return type, the rest looks solid

featureflags.go
dustinbyrne: There's other code paths in this function which also return `nil` on error.
https://github.com/PostHog/posthog-go/blob/e529d35d7cef1c374bd2519b8a4ec75b62730bdd/featureflags.go#L614-L627

What'd I'd observed was that in certain cases it'd be possible to get `false, Error{}` back from `getFeatureFlags`, so this change was to normalize that response to always be `nil, Error{}` if it was inconclusive.

Though, I do want to circle back on this before merging. I did rebase before pushing and it's possible that this may have changed.

PR #139, opened 2025-12-03
feat(flags): Add ETag support for local evaluation polling

haacked:
Add support for HTTP conditional requests using ETags to reduce bandwidth when polling for feature flag definitions. When flag definitions haven't changed, the server returns 304 Not Modified and the SDK skips processing.

Based on https://github.com/PostHog/posthog-python/pull/381

### DIFF
--- a/featureflags_etag_test.go
+++ b/featureflags_etag_test.go
@@ -3,23 +3,33 @@ package posthog
 import (
 	"net/http"
 	"net/http/httptest"
+	"net/url"
 	"strings"
-	"sync/atomic"
+	"sync"
 	"testing"
 	"time"
 )
 
-// waitForCondition polls until condition returns true or timeout expires.
-// Returns true if condition was met, false if timed out.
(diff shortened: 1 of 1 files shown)


featureflags.go
```diff
@@ -327,12 +328,31 @@ func (poller *FeatureFlagsPoller) run() {
...
+	currentEtag := poller.flagsEtag
+	poller.mutex.RUnlock()
+
+	res, cancel, err := poller.localEvaluationFlags(headers, currentEtag)
 	if err != nil {
 		poller.Logger.Errorf("Unable to fetch feature flags: %s", err)
 		return
 	}
 	defer cancel()
+	defer res.Body.Close()
```
haacked: Moved this call here because before:
- If status was `402` (quota limit) → returned early, body never closed (leak!) 
- If status was not `200` → returned early, body never closed (leak!)

orian (approved):

PR #174, opened 2026-03-31
fix: GetFeatureFlagPayload returns no error on no-match with OnlyEvaluateLocally

### LINKED ISSUE #76: IsFeatureEnabled and GetFeatureFlagPayload behave differently for mismatch
With `OnlyEvaluateLocally: true` set on the FeatureFlagPayload and no match for the provided payload,

IsFeatureEnabled will return `false, nil`

GetFeatureFlagPayload will return `"", errors.New("unable to compute flag locally")`

I would expect GetFeatureFlagPayload to return an empty payload and either no error or a known error, something like NoMatchError.

At least I would suggest documenting that GetFeatureFlagPayload will return an error if there is no match. The error `unable to compute flag locally` suggests that it was unable to compute the flag locally, not that it didn't match the rule.

tedre191:
## Motivation and Context

Fixes #76.

When `OnlyEvaluateLocally: true` and the flag exists but the user doesn't match, `GetFeatureFlagPayload` was returning `("", error)` with the message `"unable to compute flag locally"`. This is inconsistent with `GetFeatureFlag` and `IsFeatureEnabled` which return `(nil, nil)` and `(false, nil)` respectively in the same scenario.

The fix passes through `err` at the final return instead of always constructing a new error, matching the pattern already used by `GetFeatureFlag`.

## How did you test it?

Added regression test `TestGetFeatureFlagPayloadNoMatchReturnsNoError` using a fixture with a 0% rollout flag and `OnlyEvaluateLocally: true`, asserting `("", nil)` is returned. All existing feature flag tests continue to pass.

## Checklist

### DIFF
--- a/feature_flags_local_test.go
+++ b/feature_flags_local_test.go
@@ -323,6 +323,39 @@ func TestFeatureFlagsDontFallbackToFlagsWhenOnlyLocalEvaluationIsTrue(t *testing
 	}
 }
 
+func TestGetFeatureFlagPayloadNoMatchReturnsNoError(t *testing.T) {
+	// Regression test for: GetFeatureFlagPayload with OnlyEvaluateLocally=true should
+	// return ("", nil) on no match, consistent with IsFeatureEnabled returning (false, nil).
+	// Uses a flag with 0% rollout so it evaluates locally but never matches.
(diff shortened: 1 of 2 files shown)


tedre191:
Turns out this is dead/deprecated code, I was on an older version. Closing.

PR #179, opened 2026-04-15
feat(flags): switch local evaluation endpoint to /flags/definitions

patricio-posthog:
## Motivation and Context

The Rust feature flags definitions fleet now serves 100% of `/api/feature_flag/local_evaluation` traffic in all environments. This switches the SDK's default polling URL from the legacy Django path to the Rust endpoint's native path (`/flags/definitions`).

The old `/api/feature_flag/local_evaluation` path remains registered as a route alias on the Rust service, so older SDK versions continue to work.

## How did you test it?

- Updated all test mock HTTP servers to match the new path
- `go build ./...` passes


dustinbyrne (approved):
👍 👍

PR #191, opened 2026-04-27
feat: add EvaluateFlags() API for single-call flag evaluation

dmarticus:
## Summary

- New `Client.EvaluateFlags(payload)` returns a `*FeatureFlagEvaluations` snapshot built from a single `/flags` request. When local evaluation resolves some flags and the remote request fails, the snapshot still carries the locally-resolved flags alongside the error.
- Snapshot exposes `IsEnabled` / `GetFlag` / `GetFlagPayload` / `OnlyAccessed` / `Only` / `Keys`. Access methods fire deduped `$feature_flag_called` with full v4 metadata, the per-flag `$feature/<key>` value, and the parsed `$feature_flag_payload`. `GetFlagPayload` is silent.
- New `Capture.Flags *FeatureFlagEvaluations` attaches `$feature/<key>` and `$active_feature_flags` from the snapshot — no second `/flags` call per event. Takes precedence over `SendFeatureFlags` (with a runtime warning if both are set), and caller-supplied `Properties` override the auto-generated values on conflict.
- `SendFeatureFlags` is unchanged (no deprecations in this PR).

```go
snap, _ := client.EvaluateFlags(posthog.EvaluateFlagsPayload{DistinctId: "user-1"})
if snap.IsEnabled("new-checkout") { /* ... */ }
client.Enqueue(posthog.Capture{
    DistinctId: "user-1",
    Event:      "checkout-completed",
    Flags:      snap, // attaches $feature/<key> + $active_feature_flags, no extra /flags call
})
```

[RFC](https://github.com/PostHog/requests-for-comments-internal/pull/1020) — mirrored from [posthog-python#539](https://github.com/PostHog/posthog-python/pull/539) and [posthog-js#3476](https://github.com/PostHog/posthog-js/pull/3476), incorporating the latest review feedback from the Python PR.

## Design decisions

- **Single LRU cache for both paths**: the legacy single-flag path and the snapshot path share the same `(distinctId, key, deviceId)` dedup cache, so mixing the two APIs in one process can't double-fire `$feature_flag_called`.
- **`OnlyAccessed()` honors its name — empty by default**: returning all flags as a "safety net" when nothing was accessed contradicted the method's intent; callers who want all flags can just pass the parent snapshot directly.
- **Partial snapshot on remote failure**: if `/flags` errors after local evaluation succeeded for some flags, we return both the snapshot and the error so callers don't lose the locally-resolved work.
- **`Capture.Flags` takes precedence over `SendFeatureFlags`** with a runtime warning when both are set, and **user `Properties` override generated `$feature/<key>`** values — letting callers manually overwrite a flag's tagged value on a single event without rebuilding the snapshot.
- **Filtered children get a fresh `accessed` set**: `OnlyAccessed()` / `Only()` clones don't back-propagate access into the parent, so attaching `snap.OnlyAccessed()` to a `Capture` doesn't widen what a later `parent.OnlyAccessed()` returns.
- **Empty `DistinctId` returns a silent empty snapshot** instead of erroring — matches the Python/Node design and avoids `$feature_flag_called` events with empty distinct_ids leaking into analytics.
- **Response-level errors propagate per-flag**: `errorsWhileComputingFlags` and `quotaLimited` on the snapshot are combined comma-joined with per-flag errors (e.g. `errors_while_computing_flags,flag_missing`) in `$feature_flag_error`, matching the legacy single-flag path's granularity.
- **Filter warnings route through `Config.Logger`** rather than a dedicated config switch; users who want them silenced can pass a quiet `Logger`.
- **`flag_keys_to_evaluate` wired through the decider interface**, so existing single-flag callers pass `nil` and behavior is unchanged on the wire.

## Phase 2 follow-ups (out of scope here)

- Deprecate `Capture.SendFeatureFlags` in favor of `Capture.Flags`.
- Steer `IsFeatureEnabled` / `GetFeatureFlag` / `GetFeatureFlagResult` callers toward `EvaluateFlags` for multi-flag use.
- Plumb a `flag_definitions_loaded_at` timestamp from the local-eval poller so `$feature_flag_definitions_loaded_at` can be attached for locally-evaluated flags.
- Surface `TIMEOUT` / `CONNECTION_ERROR` / `api_error_NNN` exceptions in the snapshot's `$feature_flag_error` (currently only the response-level error categories are propagated).

> Branch was originally based on stale local `master`; force-pushed after rebasing onto `origin/main`. All tests + lint pass.

---

### DIFF
--- a/feature_flag_evaluations.go
+++ b/feature_flag_evaluations.go
@@ -1,0 +1,325 @@
+package posthog
+
+import (
+	"sort"
+	"strings"
+	"sync"
+)
+
+// FeatureFlagEvaluations is a snapshot of feature-flag evaluations for a
+// single distinct_id, returned by Client.EvaluateFlags. It is the recommended
+// way to evaluate many flags for a single user with a single network request:
+// the snapshot can power any number of IsEnabled / GetFlag checks and can be
+// attached to a Capture event via Capture.Flags so the event carries the
+// exact values the application branched on, with no additional /flags call.
+//
+// Calls to IsEnabled and GetFlag fire $feature_flag_called (deduped per
+// (distinct_id, key, device_id)). GetFlagPayload does not fire an event.
+type FeatureFlagEvaluations struct {
+	host                    featureFlagEvaluationsHost
+	distinctId              string
+	deviceId                *string
+	groups                  Groups
+	flags                   map[string]evaluatedFlagRecord
+	requestId               string
+	evaluatedAt             *int64
+	flagDefinitionsLoadedAt *int64
+	// errorsWhileComputing and quotaLimited capture response-level error
+	// signals from the /flags request. They are combined with per-flag
+	// errors in the $feature_flag_called event so consumers see the same
+	// granularity the legacy single-flag path emits.
+	errorsWhileComputing bool
+	quotaLimited         bool
+
+	mu       sync.Mutex
+	accessed map[string]struct{}
+}
+
+// evaluatedFlagRecord is the per-flag entry stored in a snapshot. All fields
+// are optional except Key and Enabled, mirroring the v4 /flags response shape
+// with an additional LocallyEvaluated marker for poller-resolved flags.
+type evaluatedFlagRecord struct {
+	Key              string
+	Enabled          bool
+	Variant          *string
+	Payload          *string
+	ID               *int
+	Version          *int
+	Reason           *string
+	LocallyEvaluated bool
+	Error            *string
+}
+
+// featureFlagEvaluationsHost is the small callback surface a snapshot uses to
+// talk back to the client. It is intentionally narrow so tests can construct
+// snapshots without spinning up a full client. Warnings are routed through
+// the SDK's Logger; users who want them silenced should pass a Logger that
+// drops Warnf calls.
+type featureFlagEvaluationsHost struct {
+	captureFlagCalledIfNeeded func(distinctId, key string, deviceId *string, properties Properties, groups Groups)
+	logger                    Logger
+}
+
+// IsEnabled reports whether the flag is enabled for this snapshot's user.
+// Unknown flags return false. The first call for a given key fires
+// $feature_flag_called with full evaluation metadata; subsequent calls with
+// the same response are deduped against the client's per-distinct_id cache.
+func (e *FeatureFlagEvaluations) IsEnabled(key string) bool {
+	if e == nil {
+		return false
+	}
+	flag, ok := e.flags[key]
+	e.recordAccess(key)
+	if !ok {
+		return false
+	}
+	return flag.Enabled
+}
+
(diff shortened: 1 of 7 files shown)


config.go
```diff
@@ -128,6 +128,13 @@ type Config struct {
 	// Defaults to CompressionNone (no compression).
 	Compression CompressionMode
 
+	// FeatureFlagsLogWarnings controls whether the FeatureFlagEvaluations
+	// filter helpers (OnlyAccessed and Only) emit warnings when called in
+	// unexpected ways — for example, calling OnlyAccessed before any flag has
+	// been accessed, or passing unknown keys to Only. The zero value (nil)
+	// means warnings are emitted; set to Ptr(false) to silence them.
+	FeatureFlagsLogWarnings *bool
```
dustinbyrne: same as https://github.com/PostHog/posthog-ruby/pull/137#discussion_r3170165738

feature_flag_evaluations.go
```diff
@@ -0,0 +1,319 @@
...
+	case flag.Variant != nil:
+		response = *flag.Variant
+	default:
+		response = true
+	}
+
+	properties := NewProperties().
+		Set("$feature_flag", key).
+		Set("$feature_flag_response", response).
+		Set("locally_evaluated", found && flag.LocallyEvaluated)
```
dustinbyrne, line 242: The other SDKs were setting `$feature/<key>` as well

dustinbyrne: \+ `$feature_flag_payload`

posthog.go
```diff
@@ -699,6 +728,234 @@ func (c *client) getAllFlagsWithContext(ctx context.Context, flagConfig FeatureF
...
+			payload.DeviceId,
+			payload.Groups,
+			payload.PersonProperties,
+			payload.GroupProperties,
+			disableGeoIP,
+			payload.FlagKeys,
+		)
+		if err != nil {
+			return nil, err
+		}
```
dustinbyrne: Similar to https://github.com/PostHog/posthog-rs/pull/105#discussion_r3170437174

posthog.go
```diff
@@ -365,7 +373,15 @@ func (c *client) Enqueue(msg Message) (err error) {
...
 		m.Timestamp = makeTimestamp(m.Timestamp, ts)
-		if m.shouldSendFeatureFlags() {
+		if m.Flags != nil {
+			if m.shouldSendFeatureFlags() {
+				c.Warnf("[FEATURE FLAGS] Both Flags and SendFeatureFlags were set on Capture; using Flags and ignoring SendFeatureFlags.")
+			}
+			if m.Properties == nil {
+				m.Properties = NewProperties()
+			}
+			m.Properties.Merge(m.Flags.eventProperties())
```
dustinbyrne: nit: the Python SDK merged generated flag properties first, letting user properties override conflicts

dustinbyrne (approved):

PR #193, opened 2026-04-30
Fix no-op client for empty API key

marandaneto:

This PR makes missing API key handling safe and consistent:

- If the project API key is missing or blank after trimming whitespace, the SDK returns a disabled `noopClient` and all public methods become no-ops.
- `noopClient` performs no validation and returns zero/non-error values.
- Methods that depend on `PersonalApiKey` now no-op when it is missing or blank after trimming whitespace, log a warning, and return empty/disabled values instead of errors.
- Remote `/flags` evaluation still works when a project API key exists but no `PersonalApiKey` is configured.
- Reuses shared empty/no-op feature flag result instances for no-op early returns.
- Renames `noop.go` to `noop_posthog.go`.
- Removes the now-unused `ErrNoPersonalAPIKey` error and updates method comments to document no-op behavior.

- `go test ./... -run 'TestNewWithConfig_LogsErrorForBlankAPIKeyAfterTrim|TestNew_BlankAPIKeyReturnsNoopClient|TestNewWithConfig_BlankAPIKeyReturnsNoopClientWithoutRequests|TestNewWithConfig_BlankAPIKeyWithPersonalAPIKeyReturnsNoopClient|TestFeatureFlagsWithNoPersonalApiKey|TestGetFeatureFlagPayloadWithNoPersonalApiKey|TestGetFeatureFlagWithNoPersonalApiKey|TestGetAllFeatureFlagsWithNoPersonalApiKey|TestClient_GetRemoteConfigPayload_IncludesTokenParameter'`
- `go test ./... -run '^$'`

Note: full `go test ./...` still fails on existing `$os_distro` fixture mismatches unrelated to this change.

## :pencil: Checklist

### If releasing new changes


posthog.go
```diff
@@ -679,14 +689,13 @@ func (c *client) GetRemoteConfigPayload(flagKey string) (string, error) {
...
+// If no PersonalApiKey is configured, this is a no-op and returns an empty slice without an error.
+// Not to be confused with GetAllFlags, which returns all flags and their values for a given user.
 func (c *client) GetFeatureFlags() ([]FeatureFlag, error) {
 	if c.featureFlagsPoller == nil {
-		err := fmt.Errorf("cannot use feature flags: %w", ErrNoPersonalAPIKey)
-		c.Logger.Debugf(err.Error())
-		return nil, err
+		c.warnPersonalAPIKeyNoop("GetFeatureFlags")
+		return emptyFeatureFlags, nil
 	}
```
dustinbyrne, line 699: I'm kind of torn on us porting this behavior to Golang because it's idiomatic to return the error here

Though, the net effect is essentially unchanged as it was also a no-op before

posthog.go
```diff
@@ -203,6 +209,7 @@ func NewWithConfig(apiKey string, config Config) (cli Client, err error) {
 	apiKey = strings.TrimSpace(apiKey)
 	if len(apiKey) == 0 {
 		config.Logger.Errorf("posthog apiKey is empty after trimming whitespace; check your project API key")
+		return newNoopClient(config), nil
```
dustinbyrne, line 212: Strongly suggest `return nil, err` and avoiding the noop client

dustinbyrne (approved):
I'm team `nil, err`, but will leave it up to you. There's some good stuff in here regardless, like returning early from the client constructor.

posthog.go
```diff
@@ -471,9 +478,8 @@ func (c *client) IsFeatureEnabled(flagConfig FeatureFlagPayload) (interface{}, e
 
 func (c *client) ReloadFeatureFlags() error {
 	if c.featureFlagsPoller == nil {
-		err := fmt.Errorf("cannot use feature flags: %w", ErrNoPersonalAPIKey)
```
marandaneto, line 474: not sure if we should kill `ErrNoPersonalAPIKey` here since the public method already returns an `error` type and panic is different than crashing the hosting app
wdyt @dustinbyrne ?

marandaneto: not very xp with go panics

posthog.go
marandaneto: oh just read this, see https://github.com/PostHog/posthog-go/pull/193/changes#r3203757089
100% agree, will fix

posthog.go
marandaneto: i dont agree because this will lead to a breaking change

```
     client := posthog.New("")
     client.Enqueue(...) // panic: nil pointer/interface method call
```

which is exactly what we dont want to do
client can be nil because its an interface but our docs and example never mentioned that client can be nil, so it might be the case that people is never checking for that

marandaneto:
@dustinbyrne can you double check? i changed a few things to keep the go errors ergonomics

marandaneto:
i dont like mixing `ErrNoDistinctID` and `ConfigError` with `DistinctId required` as reason but replacing would be a breaking change, and using `DistinctId required` where `ErrNoDistinctID` is thrown now is strange because its not a config error when calling `NewWithConfig`, ideally they would share the same error, like all using `ErrNoDistinctID` or all using `ConfigError` but with a different name than "Config"

noop_posthog.go
```diff
@@ -0,0 +1,67 @@
...
+	noopFeatureFlagResult      = &FeatureFlagResult{Enabled: false}
+	noopFeatureFlagEvaluations = &FeatureFlagEvaluations{flags: emptyEvaluatedFlagRecords}
+)
+
+func newNoopClient(config Config) Client {
+	return &noopClient{Config: config}
+}
+
+func (c *noopClient) Enqueue(Message) error {
+	return nil
```
dustinbyrne: Since we're keeping this, consider returning a static error on all of these. We can keep the default values too, but otherwise there's no way to know that it's in a bad state.

posthog.go
dustinbyrne: I'm not sure what you mean exactly, but the change here looks good to me.

Most calls in golang end up returning an error type, either by itself or in a tuple. It's nice, because it means the caller needs to explicitly do something with the error, either read it or ignore it.

Returning an error can't cause a panic, but if the caller _chooses to ignore_ an error it could lead to issues.

posthog.go
dustinbyrne: Yeah I was just looking at `NewWithConfig`, I didn't notice `New`
```go
func New(apiKey string) Client {
	// Here we can ignore the error because the default config is always valid.
	c, _ := NewWithConfig(apiKey, Config{})
	return c
}
```



so fixing would be a major change, `noopClient` it is!

noop_posthog.go
marandaneto: good point, done https://github.com/PostHog/posthog-go/pull/193/changes/0cb7c40dfdb86effb1cdd44daaf70bda86edbc17

noop_posthog.go
```diff
@@ -19,49 +19,49 @@ func newNoopClient(config Config) Client {
 }
 
 func (c *noopClient) Enqueue(Message) error {
-	return nil
+	return ErrSDKDisabled
 }
 
 func (c *noopClient) IsFeatureEnabled(FeatureFlagPayload) (interface{}, error) {
-	return false, nil
+	return false, ErrSDKDisabled
```
dustinbyrne, line 26: nice

PR #200, opened 2026-05-20
fix: reject leading-zero semver values in local evaluation

dmarticus:
## Summary

Per [semver 2.0.0 §2](https://semver.org/#spec-item-2), numeric identifiers MUST NOT include leading zeros. Values like `1.07.3` or `01.2.3` are not valid semver — the local feature flag evaluator currently parses them silently (via `strconv.Atoi("07") → 7`), which means a person property of `1.07.3` would incorrectly match a `semver_eq` condition against `1.7.3`.

This PR makes the parser reject leading zeros in numeric identifiers. Both override values and flag values are validated; invalid inputs surface an `InconclusiveMatchError` so the condition does not match.

## Changes

- New `parseSemverNumeric` helper that wraps `strconv.Atoi` with a leading-zero check.
- `parseSemver` and `computeWildcardBounds` use the new helper.
- Inverted the existing `parseSemver` subtest that asserted `01.02.03` parses to `{1, 2, 3}` — it now asserts an error.
- Added matchProperty-level rejection tests for override values and for flag values across `semver_caret`, `semver_tilde`, `semver_wildcard`, and `semver_gt`.

## Test plan

- [x] `go test -run "Semver|TestParseSemver|TestMatchPropertySemver" .` — all pass
- [x] New rejection subtests pass
- [x] `gofmt -l .` clean

### DIFF
--- a/.changeset/strict-semver-leading-zeros.md
+++ b/.changeset/strict-semver-leading-zeros.md
@@ -1,0 +1,5 @@
+---
+"posthog-go": patch
+---
+
+Reject semver values with leading zeros in local flag evaluation. Per semver 2.0.0 §2, numeric identifiers must not include leading zeros — values like `1.07.3` are not valid semver and should not match targeting conditions. Both override values and flag values are now validated; invalid inputs surface an `InconclusiveMatchError` so the condition does not match.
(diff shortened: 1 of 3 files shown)


marandaneto (approved):

marandaneto:
missing changelog entry @dmarticus

PR #201, opened 2026-05-20
feat: add server-side request context

dustinbyrne:

Adds server-side request context support for Go `net/http` apps, matching the request-context patterns used in other PostHog server SDKs.

This lets handlers enqueue capture and exception events with request-scoped metadata and optional frontend tracing headers, while keeping explicit event identity authoritative and falling back to personless events when no identity is available.

Key changes:
- Add `RequestContext` helpers, `NewRequestContextMiddleware`, and `EnqueueWithContext`.
- Attach safe request metadata (`$current_url`, `$request_method`, `$request_path`, `$user_agent`, `$ip`) and optional tracing header values.
- Support personless capture/exception fallback with `$process_person_profile: false`.
- Add opt-in panic capture that re-panics and preserves host-app behavior.
- Document tracing headers as client-controlled analytics context, not auth/authz.

- `gofmt -l $(git diff --name-only -- '*.go') request_context.go request_context_test.go`
- `go test ./...`

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/request_context.go
+++ b/request_context.go
@@ -1,10 +1,15 @@
 package posthog
 
 import (
+	"bufio"
 	"context"
 	"fmt"
+	"io"
 	"net"
 	"net/http"
+	"runtime"
+	"runtime/debug"
+	"strconv"
 	"strings"
 )
 
@@ -19,6 +24,8 @@ const (
 	propertyRequestPath          = "$request_path"
 	propertyUserAgent            = "$user_agent"
 	propertyIP                   = "$ip"
+	propertyResponseStatusCode   = "$response_status_code"
+	propertyExceptionSource      = "$exception_source"
 
 	maxRequestContextValueLength = 1000
 )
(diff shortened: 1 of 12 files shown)


dustinbyrne:
summoning greptile

request_context.go
```diff
@@ -0,0 +1,392 @@
...
+		return string(runes[:maxRequestContextValueLength])
+	}
+	return sanitized
+}
+
+func currentURL(r *http.Request) string {
+	scheme := requestScheme(r)
+	host := requestHost(r)
+	if scheme == "" || host == "" {
+		return ""
```
marandaneto, line 352: for this all and other methods here better to return nil instead of empty if we cannot derive the prop

marandaneto (approved):

request_context.go
dustinbyrne: The props are all ultimately fed in to `addPropertyIfPresent` which explicitly tests the empty case. Switching the return type to `*string` feels like it introduces dereference noise for the same effect. I thought that maybe returning a `(string, bool)` would be a decent middle ground, but it also felt redundant given the later calls to `addPropertyIfPresent`. I've added a comment to clarify the intent. This is all purely internal, so we're not locked in by any means.

PR #205, opened 2026-05-22
feat: add request-context panic capture

dustinbyrne:

Builds on the server-side request context work in #201 by adding opt-in panic capture for `net/http` middleware.

This lets apps capture panics as PostHog exception events with request metadata and response status, while still re-panicking with the original value so host application behavior is preserved.

Key changes:
- Add `WithCapturePanics(client)` request-context middleware option.
- Capture panic exceptions with `$response_status_code` and `$exception_source: "panic"`.
- Preserve optional `http.ResponseWriter` interfaces only when the wrapped writer supports them.
- Ensure panic-capture failures cannot replace the original application panic.

- `go test ./...`

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/.changeset/bright-ravens-panic.md
+++ b/.changeset/bright-ravens-panic.md
@@ -1,0 +1,5 @@
+---
+"posthog-go": minor
+---
+
+Add opt-in panic capture for request context middleware.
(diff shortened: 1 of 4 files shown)


dustinbyrne:
summoning greptile

marandaneto (approved):

PR #215, opened 2026-06-02
feat: emit $is_server property on captured events

turnipdabeets:
Adds `$is_server: true` to every event captured by this SDK so PostHog ingestion can identify server-side events.


.changeset/is-server-property.md
```diff
@@ -0,0 +1,5 @@
+---
+"posthog-go": patch
+---
+
+Emit `$is_server` on identify and group-identify events too (was only on capture); server-side events are now consistently identifiable.
```
marandaneto: looks like wrong. AI slop?

marandaneto:
if we decide to add `$is_server` to all SDKs, there should be a way to turn it off. Some SDKs can also be used as a client, e.g., the Go SDK could be a CLI and wouldn't be a server-side SDK
same for java, node, etc

.changeset/is-server-property.md
turnipdabeets: updated

turnipdabeets:
> if we decide to add `$is_server` to all SDKs, there should be a way to turn it off. Some SDKs can also be used as a client, e.g., the Go SDK could be a CLI and wouldn't be a server-side SDK same for java, node, etc

Made these configurable now, but defaulted to true.

capture_test.go
```diff
@@ -52,3 +55,40 @@ func TestCaptureValidWithDistinctId(t *testing.T) {
...
+	expectKeys := map[string]interface{}{
+		"$lib":       SDKName,
+		"$is_server": true,
+	}
+	for k, want := range expectKeys {
+		if got := props[k]; got != want {
+			t.Errorf("property %q: expected %v (%T), got %v (%T)", k, want, want, got, got)
+		}
+	}
+}
```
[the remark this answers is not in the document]
turnipdabeets: Simplified the map-based capture assertions to read apiMsg.Properties directly. Kept the JSON round-trip on the alias/exception tests, where $is_server is a struct field with json omitempty and the round-trip verifies the actual serialized wire output. Thanks!

eli-r-ph (approved):

PR #217, opened 2026-06-04
feat(feature-flags): support early_exit in local evaluation

gustavohstrassburger:

PostHog added an "early exit" option to feature flags server-side (the Rust evaluation engine). When a flag's `filters.early_exit` is `true`, local condition evaluation must STOP and return a definitive disabled result as soon as a condition group's property filters match (or the group has no property filters) but the rollout percentage EXCLUDES the user — instead of falling through to evaluate later condition groups.

This was already ported to posthog-node and posthog-python. This PR ports the same behavior to the Go SDK's local evaluation.

### What changed

- Added an `EarlyExit bool` field (JSON tag `early_exit`) to the `Filter` struct.
- Refactored `isConditionMatch` to return a tri-state result (`conditionMatch` / `conditionNoMatch` / `conditionOutOfRolloutBound`) instead of a bare bool, so the rollout-exclusion path is distinguishable from a property mismatch. This mirrors the Python port's `MATCH` / `NO_MATCH` / `OUT_OF_ROLLOUT_BOUND` enum.
- In `matchFeatureFlagProperties`, when a group's outcome is `conditionOutOfRolloutBound` AND `filters.early_exit` is enabled, evaluation returns a definitive `false` immediately. Property mismatches (`conditionNoMatch`) always fall through, even with early_exit on. Behaviour is unchanged when `early_exit` is absent or `false`.

The change is intentionally minimal and focused on `early_exit`; no dependencies, workflows, or CODEOWNERS were touched.

Added a table-driven test `TestFeatureFlagEarlyExit` (in `feature_flags_local_test.go`) backed by four local-evaluation fixtures. Each fixture defines two condition groups that both match on `region=USA`: the first with `rollout_percentage: 0` (always out of rollout bound) and the second with `rollout_percentage: 100` (always in rollout). The cases cover:

1. `early_exit: true` → returns `false` WITHOUT evaluating the later group that would match.
2. `early_exit` unset → falls through to the later matching group (regression, `true`).
3. `early_exit: false` → falls through (`true`).
4. `early_exit: true` but the first group fails on a PROPERTY filter (not rollout) → does NOT early-exit, falls through to the matching group (`true`).

`go build ./...`, `go vet ./...`, `gofmt -l` (clean), and the full `go test ./...` suite all pass locally.

## :pencil: Checklist

### If releasing new changes

- [x] Added a `## Unreleased` entry to `CHANGELOG.md` describing the change.

---

### DIFF
--- a/featureflags.go
+++ b/featureflags.go
@@ -1040,8 +1040,10 @@ func (poller *FeatureFlagsPoller) matchFeatureFlagProperties(
 			// The property filters matched (or there were none) but the rollout
 			// excluded the user. When early_exit is enabled, mirror the server
 			// engine by returning a definitive disabled result instead of
-			// falling through to later condition groups.
-			if flag.Filters.EarlyExit {
+			// falling through to later condition groups. An inconclusive prior
+			// group takes priority: we can't decide locally, so we must fall
+			// back to the server.
+			if flag.Filters.EarlyExit && !isInconclusive {
 				return false, nil
 			}
 		case conditionNoMatch:
--- a/featureflags.go
+++ b/featureflags.go
@@ -1040,10 +1040,13 @@ func (poller *FeatureFlagsPoller) matchFeatureFlagProperties(
 			// The property filters matched (or there were none) but the rollout
 			// excluded the user. When early_exit is enabled, mirror the server
 			// engine by returning a definitive disabled result instead of
-			// falling through to later condition groups. An inconclusive prior
-			// group takes priority: we can't decide locally, so we must fall
-			// back to the server.
-			if flag.Filters.EarlyExit && !isInconclusive {
+			// falling through to later condition groups. If a prior group was
+			// inconclusive we can't decide locally, so return errInconclusiveMatch
+			// to force a server fallback rather than continuing evaluation.
+			if flag.Filters.EarlyExit {
+				if isInconclusive {
+					return false, errInconclusiveMatch
+				}
 				return false, nil
 			}
 		case conditionNoMatch:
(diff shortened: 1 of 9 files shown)


dustinbyrne (commented):
Validated the early_exit edge case locally; leaving one blocking correctness note inline.

featureflags.go
```diff
@@ -1030,6 +1036,19 @@ func (poller *FeatureFlagsPoller) matchFeatureFlagProperties(
...
 				return getMatchingVariant(flag, effectiveBucketingId), nil
 			}
+		case conditionOutOfRolloutBound:
+			// The property filters matched (or there were none) but the rollout
+			// excluded the user. When early_exit is enabled, mirror the server
+			// engine by returning a definitive disabled result instead of
+			// falling through to later condition groups. An inconclusive prior
+			// group takes priority: we can't decide locally, so we must fall
+			// back to the server.
+			if flag.Filters.EarlyExit && !isInconclusive {
```
dustinbyrne: Blocking correctness issue: this still masks a prior inconclusive condition.

If an earlier condition set `isInconclusive = true`, this guard prevents the immediate `false`, but evaluation then continues. A later condition group can still match and return `true`, bypassing the final `errInconclusiveMatch` fallback.

Repro shape:
1. group 1 is inconclusive locally, e.g. missing group properties
2. group 2 is `OutOfRolloutBound` with `early_exit=true`
3. group 3 matches

Current result: `true`
Expected: inconclusive / server fallback

When `early_exit && conditionOutOfRolloutBound && isInconclusive`, we should return `errInconclusiveMatch` immediately instead of continuing.

dustinbyrne (approved):
nice

dustinbyrne (approved):
pls take a look at https://github.com/PostHog/posthog-go/blob/main/RELEASING.md before merge

PR #220, opened 2026-06-12
ci: add public API snapshot check

marandaneto:
Add automated coverage for public API signature changes so pull requests surface intentional SDK API changes during review without flagging documentation-only edits.

- [x] Ran `make api-update`
- [x] Ran `make api-diff`

## :pencil: Checklist

### If releasing new changes


.github/workflows/unit-tests.yml
```diff
@@ -35,6 +35,22 @@ jobs:
       - name: Build
         run: go build .
 
+  public-api:
+    name: Public API
+    if: github.event_name == 'pull_request'
```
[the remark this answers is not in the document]
marandaneto: Fixed in 15a0573 — removed the pull_request-only condition so the Public API job also runs for merge_group events.

api/public-api.txt
```diff
@@ -0,0 +1,1717 @@
...
+
+## github.com/posthog/posthog-go
+
+package posthog // import "github.com/posthog/posthog-go"
+
+
+CONSTANTS
+
+const (
+	// SDKName is the library identifier sent in event metadata.
```
ioannisj: should doc comments be part of the api we generate and check? Not a blocker but this will trigger when just the docs change

ioannisj (approved):

api/public-api.txt
marandaneto: nope, just signature, will check

marandaneto: Good call. Fixed in d868db7 — the generated public API snapshot now strips doc comments/doc text, so documentation-only changes won’t affect the API check.

marandaneto:
https://github.com/PostHog/posthog-go/pull/222#issuecomment-4707451894

PR #221, opened 2026-06-12
feat: add native stack trace extractor for server-side symbolication

cat-ph:
## Problem

Today the SDK resolves stack traces in-process via `runtime.CallersFrames` and sends `platform: "go"` frames. That gives function/file/line, but no source context in the UI, and inlined calls surface however the local runtime expands them.

PostHog now supports server-side native symbolication (same pipeline as Apple/Rust): the SDK sends raw instruction addresses plus the binary's identity (`$debug_images`), and the server resolves them against debug symbols uploaded with `posthog-cli symbol-sets upload` — exact inline expansion, source context, and symbolication that keeps working for stripped binaries.

## Changes

Opt-in native capture (the default extractor is unchanged — `platform: "native"` frames require a PostHog version that understands them, so older self-hosted instances should stay on the default):

- `NativeStackTraceExtractor` — emits frames with raw `instruction_addr` (plus `symbol_addr`/`image_addr`), keeping the runtime-resolved function/file/line as a fallback for when no symbols are uploaded.
- `NewNativeException` — `NewDefaultException` counterpart wired to the native extractor and debug images.
- `DebugImage` + `GetDebugImages()` — reports the main executable: ELF on Linux (GNU build id from the mapped `/proc/self/exe`, range from `/proc/self/maps`), Mach-O on macOS (LC_UUID; ASLR slide computed via `debug/gosym` over `__gopclntab`, since the darwin symtab no longer carries Go symbols). Other platforms degrade to address-only frames.
- `DebugImageProvider` — optional interface so integrations that take a `StackTraceExtractor` can attach images; the slog capture handler now does.
- `StackFrame` gains `lang`, `instruction_addr`, `symbol_addr`, `image_addr` (all omitempty); `Exception`/`ExceptionInApiProperties` gain `$debug_images`.

Two findings baked into the implementation, worth reviewer attention:

- **Raw pcs, not `frame.PC`**: `runtime.CallersFrames` pre-decrements the pc (`frame.PC == pc-1`); the server also subtracts 1 to target the call instruction (shared contract with the other native SDKs). Sending `frame.PC` would double-adjust, so we send the raw `runtime.Callers` value.
- **Inline collapse**: `runtime.Callers` synthesizes separate pcs entries for inlined calls *and their parents* within one physical frame. The server re-derives the full inline chain from the innermost address, so the extractor emits one frame per physical frame (the innermost entry) and swallows the synthesized parents — otherwise every inlined call's parent would appear twice after symbolication. `TestNativeExtractorCollapsesInlineFrames` exercises this against real compiler inlining.

Linux requires linking with `-ldflags "-B gobuildid"` (Go adds no GNU build id by default); documented on the relevant types and in the changeset.

## How did you test this code?

- `go test -race ./...` — including new tests for the build-id → debug-id derivation (vector cross-checked against symbolic), wire frame order, inline collapse, debug-image extent/arch invariants, `$debug_images` serialization, and slog handler attachment.
- Local E2E against a dev PostHog: built a linux/amd64 binary with `-B gobuildid`, uploaded symbols via `posthog-cli symbol-sets upload`, captured an exception through this extractor, and verified the server resolves the frames (including exact inline expansion) in the error tracking UI. Server-side counterpart: PostHog/posthog#62586 / PostHog/posthog#62590 / PostHog/posthog#62592 plus PostHog/posthog#63348 stacked on them.


error_tracking_native_test.go
```diff
@@ -0,0 +1,240 @@
...
+
+	// Short build ids are zero-padded to 16 bytes
+	if got, want := debugIDFromGNUBuildID([]byte{0xab, 0xcd}), "0000cdab-0000-0000-0000-000000000000"; got != want {
+		t.Errorf("debugIDFromGNUBuildID(short) = %q, want %q", got, want)
+	}
+
+	if got := debugIDFromGNUBuildID(nil); got != "" {
+		t.Errorf("debugIDFromGNUBuildID(nil) = %q, want empty", got)
+	}
+}
```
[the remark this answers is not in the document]
cat-ph: Keeping these as inline assertions: three cases with distinct shapes (reference vector with a documented provenance, padding behavior, nil guard) read clearer flat than table-driven here, and the Go repo does not use a table-driven convention for such micro-helpers.

error_tracking_native_linux.go
```diff
@@ -0,0 +1,148 @@
...
+		name := data[:nameSize]
+		desc := data[alignedName : alignedName+descSize]
+		data = data[alignedName+alignedDesc:]
+
+		if noteType == gnuBuildIDType && bytes.Equal(name, []byte("GNU\x00")) {
+			return desc
+		}
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
cat-ph: Fixed in bc7f86d - note header fields now use file.ByteOrder; cross-checked with CGO_ENABLED=0 GOOS=linux GOARCH=s390x go build. (The little-endian GUID shuffle in debugIDFromGNUBuildID is unchanged on purpose: it mirrors how symbolic derives debug ids on the little-endian hosts PostHog runs on.)

api/public-api.txt
```diff
@@ -317,6 +317,21 @@ type ConfigError struct {
...
+	ImageAddr string `json:"image_addr"`
+	ImageSize uint64 `json:"image_size,omitempty"`
+	ImageVmaddr string `json:"image_vmaddr,omitempty"`
+	CodeFile string `json:"code_file,omitempty"`
+	Arch string `json:"arch"`
+}
+
+type DebugImageProvider interface {
+	GetDebugImages() []DebugImage
+}
```
marandaneto, line 333: i think this can be private

api/public-api.txt
```diff
@@ -384,6 +400,11 @@ func NewDefaultException(
 ) Exception
 
 
+func NewNativeException(
+	timestamp time.Time,
+	distinctID, title, description string,
+) Exception
```
marandaneto, line 406: i think this can be private

api/public-api.txt
```diff
@@ -708,6 +730,15 @@ type Message interface {
...
 
 
+type NativeStackTraceExtractor struct {
+	InAppDecider InAppDecider
+}
+
+
+func (n NativeStackTraceExtractor) GetDebugImages() []DebugImage
+
+func (n NativeStackTraceExtractor) GetStackTrace(skip int) *ExceptionStacktrace
```
marandaneto, line 740: same here

error_tracking_native_other.go
```diff
@@ -0,0 +1,10 @@
+//go:build !linux && !darwin
+
+package posthog
+
+// loadMainImage is unsupported on this platform: frames still carry raw
+// addresses, but no debug image is reported, so they fall back to the
+// runtime-resolved function, file, and line.
+func loadMainImage() mainImageInfo {
+	return mainImageInfo{}
+}
```
marandaneto, line 10: is that windows as well?
we'd need PE/CodeView/PDB metadata

marandaneto:
TBC (pls check): If frames come from cgo/shared libraries/plugins, they won’t get matching $debug_images

marandaneto:
sorry miscliked the close with the comment

api/public-api.txt
```diff
@@ -317,6 +317,21 @@ type ConfigError struct {
 
 func (e ConfigError) Error() string
 
+type DebugImage struct {
```
marandaneto, line 320: dont we need `debug_file`?

error_tracking_native_other.go
cat-ph: yea that _excludes_ windows atm, kinda intentionally out of scope here, I wanna revisit (rust/go/maybe dotnet) after getting ELF and mach-o working + symbolication

api/public-api.txt
cat-ph: shouldn't need to, cymbal doesn't have a `debug_file` field, we match them by the debug_id; debug_file would be if we had a separate debug-info file (PDB I think), for DWARF it's embedded in the binary

cat-ph:
thanks for looking @marandaneto and sorry, I'll put this back to draft for a bit, trying to simplify the approach - I think go *always* resolves/symbolicates client-side anyway, so we could simplify and send this with the default stacktrace extractor all the time (instead of a separate native path), with the debug IDs and then only enrich the source context

there's also the rust case where we have debug symbols and can resolve client-side, I'll try to update cymbal to use client-side resolution when that exists and only enrich source context (if available through uploaded symbol sets)

error_tracking_native_other.go
marandaneto: no prob, lets jsut add a note to the changelog to be clear (no win support)

marandaneto:
> thanks for looking @marandaneto and sorry, I'll put this back to draft for a bit, trying to simplify the approach - I think go _always_ resolves/symbolicates client-side anyway, so we could simplify and send this with the default stacktrace extractor all the time (instead of a separate native path), with the debug IDs and then only enrich the source context
> 
> there's also the rust case where we have debug symbols and can resolve client-side, I'll try to update cymbal to use client-side resolution when that exists and only enrich source context (if available through uploaded symbol sets)

no need to be sorry, thats how we figure stuff and move on!

PR #224, opened 2026-06-17
feat: add before send hook

### LINKED ISSUE #135: Support `before_send` hook like JS/Py/Ruby SDK
Something like this:

```go
	client, err := posthog.NewWithConfig(
		config.Posthog.APIKey,
		posthog.Config{},
	)

	if err != nil {
		log.Fatal("failed to initialize", "error", err)
	}

	client.BeforeSend(func(msg posthog.Message) posthog.Message {
		switch m := msg.(type) {
		case posthog.Exception:
			// would be nice if we can access the actual exception here,
			// instead of just the already transformed data
			return m
		default:
			return m
		}
	})
	client.BeforeSend(func(msg posthog.Message) posthog.Message {
		// is not sent if `nil` is returned
		return nil
	})
	client.BeforeSend(func(msg posthog.Message) posthog.Message {
		// never runs
	})
```

marandaneto:

Adds support for a `BeforeSend` hook so users can inspect, modify, or drop messages after SDK enrichment and before they are serialized and sent to PostHog.

Fixes https://github.com/PostHog/posthog-go/issues/135.

- `go test ./...`
- `make api-diff`

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Implemented with pi. The issue was investigated against the current SDK first, then `Config.BeforeSend` was added for construction-time hook registration. The hook runs after enrichment and before `APIfy()`, so callers receive typed SDK messages such as `posthog.Exception`; returning nil drops the message, and panic/invalid/type-changing hooks are logged and dropped.

Follow-up commits updated the public API snapshot after CI reported the intentional exported API additions, addressed review feedback by parameterizing BeforeSend tests, simplifying helper names, refining the API to construction-time registration only, and simplifying hook execution by passing the enriched message directly without cloning.

### DIFF
--- a/alias.go
+++ b/alias.go
@@ -11,6 +11,7 @@ var _ Message = (*Alias)(nil)
 // Timestamp, and DisableGeoIP, then sends the message as a $create_alias event.
 type Alias struct {
 	// Type is reserved for SDK serialization and is overwritten by Enqueue.
+	// Deprecated: this field is ignored by PostHog ingestion and is kept for backwards compatibility.
 	Type string
 	// Uuid is an optional event UUID. If empty, Enqueue generates a random UUID.
 	Uuid string
@@ -58,7 +59,8 @@ type AliasInApiProperties struct {
 
 // AliasInApi is the wire-format payload produced from an Alias message.
 type AliasInApi struct {
-	// Type is the message type sent to the batch API.
+	// Type is the legacy message type sent to the batch API.
+	// Deprecated: this field is ignored by PostHog ingestion and is kept for backwards compatibility.
 	Type string `json:"type"`
 	// Uuid is the event UUID sent to the batch API.
 	Uuid string `json:"uuid"`
--- a/before_send_test.go
+++ b/before_send_test.go
@@ -1,9 +1,11 @@
 package posthog
 
 import (
+	"fmt"
 	"io"
 	"net/http"
 	"net/http/httptest"
+	"strings"
 	"sync/atomic"
 	"testing"
 	"time"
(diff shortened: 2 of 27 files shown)


before_send_test.go
```diff
@@ -0,0 +1,204 @@
...
+
+func firstBeforeSendProperties(t *testing.T, batch map[string]interface{}) map[string]interface{} {
+	t.Helper()
+
+	message := firstBeforeSendMessage(t, batch)
+	properties, ok := message["properties"].(map[string]interface{})
+	require.True(t, ok)
+	return properties
+}
+
```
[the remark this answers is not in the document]
marandaneto: Fixed in 84e61a2 — renamed the batch test helpers to readBatch, firstMessage, and firstProperties.

posthog.go
```diff
@@ -299,6 +310,84 @@ func makeHttpClient(transport http.RoundTripper, timeout time.Duration) http.Cli
...
 }
 
+func cloneBeforeSendProperties(properties Properties) Properties {
+	if properties == nil {
+		return nil
+	}
+	clone := make(Properties, len(properties))
+	for key, value := range properties {
+		clone[key] = value
+	}
```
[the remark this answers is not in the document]
marandaneto: Fixed in 84e61a2 — renamed the private clone helpers to cloneProperties, cloneGroups, and cloneExceptionList.

posthog.go
```diff
@@ -299,6 +310,84 @@ func makeHttpClient(transport http.RoundTripper, timeout time.Duration) http.Cli
...
+}
+
+func cloneMessage(msg Message) Message {
+	switch m := dereferenceMessage(msg).(type) {
+	case Alias:
+		return m
+	case Identify:
+		m.Properties = cloneBeforeSendProperties(m.Properties)
+		return m
+	case GroupIdentify:
```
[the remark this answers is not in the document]
marandaneto: Fixed in 84e61a2 — Capture clones passed to hooks now clear Flags and SendFeatureFlags after their effects have already been expanded into properties, and a regression test covers this.

before_send_test.go
```diff
@@ -0,0 +1,204 @@
...
+	}))
+
+	message := firstBeforeSendMessage(t, readBeforeSendBatch(t, body))
+	properties, ok := message["properties"].(map[string]interface{})
+	require.True(t, ok)
+	require.Equal(t, "user-123", message["distinct_id"])
+	require.Equal(t, true, properties["after_error"])
+	require.Nil(t, properties["panic_leaked"])
+	require.Nil(t, properties["invalid_leaked"])
+}
```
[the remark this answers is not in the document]
marandaneto: Fixed in 84e61a2 — parameterized the BeforeSend capture hook scenarios with t.Run table cases, including separate panic and invalid-return coverage.

alias.go
```diff
@@ -11,6 +11,7 @@ var _ Message = (*Alias)(nil)
 // Timestamp, and DisableGeoIP, then sends the message as a $create_alias event.
 type Alias struct {
 	// Type is reserved for SDK serialization and is overwritten by Enqueue.
+	// Deprecated: this field is ignored by PostHog ingestion and is kept for backwards compatibility.
```
marandaneto, line 14: our data structure does not have a high level `type` afaik

posthog.go
```diff
@@ -390,6 +437,11 @@ func (c *client) EnqueueWithContext(ctx context.Context, msg Message) (err error
 		m.Timestamp = makeTimestamp(m.Timestamp, ts)
 		m.DisableGeoIP = c.GetDisableGeoIP()
 		m.IsServer = c.GetIsServer()
+		processed, shouldSend := c.processBeforeSend(m)
```
ioannisj, line 585: Bot says that maps in go are reference types, so m.Properties is still points to the original properties so there's a data race condition here. we'll probably need to clone to a private map

message.go
```diff
@@ -25,6 +25,10 @@ type Callback interface {
 	Failure(APIMessage, error)
 }
 
+// BeforeSendFunc is called before a message is converted to the PostHog API format.
+// It can return a modified message, or nil to drop the message.
```
ioannisj: Should we also mention here that panic, invalid and returning a different type will also drop the message?

posthog.go
marandaneto: Fixed in 3ff50e8 — BeforeSend now receives isolated map fields for message types with mutable maps, so hook mutations do not race with or mutate the caller-owned Properties/Groups maps. Added TestBeforeSendDoesNotMutateOriginalProperties as coverage.

message.go
marandaneto: Fixed in 3ff50e8 — expanded the BeforeSendFunc doc comment to state that nil, panic, invalid messages, and different message types all drop the message.

alias.go
marandaneto: Confirmed — this is documenting the legacy SDK field precisely because ingestion does not consume a meaningful high-level type here. I kept the deprecation/no-op wording so callers know not to rely on it.

posthog.go
marandaneto: it wasnt needed but better to be safe in case they use the properties concurrently in the beforesend callback

marandaneto:
CI fails becuse of https://github.com/PostHog/posthog-go/pull/226

posthog.go
```diff
@@ -299,6 +299,47 @@ func makeHttpClient(transport http.RoundTripper, timeout time.Duration) http.Cli
...
+	case Identify:
+		m.Properties = cloneMessageProperties(m.Properties)
+		return m
+	case GroupIdentify:
+		m.Properties = cloneMessageProperties(m.Properties)
+		return m
+	case Capture:
+		m.Properties = cloneMessageProperties(m.Properties)
+		m.Groups = cloneMessageGroups(m.Groups)
+		return m
```
dustinbyrne, line 437: this still exposes `Capture.Flags` / `SendFeatureFlags` as pointers to the before send hook

we should either clear or deep copy them, but i think clearing is the better option. by the time `BeforeSend` runs, flag properties have already been written to `Properties`, so there's no value in mutating these

posthog.go
```diff
@@ -299,6 +299,47 @@ func makeHttpClient(transport http.RoundTripper, timeout time.Duration) http.Cli
...
+	case Exception:
+		m.Properties = cloneMessageProperties(m.Properties)
+		return m
```
dustinbyrne, line 442: similar, we should clone `ExceptionList`, nested stack frames/mechanism pointers, and `ExceptionFingerprint`

posthog.go
dustinbyrne: it doesn't look like they're used any time after this - some mixed responsibilities going on with this struct

posthog.go
marandaneto: Fixed in b1b16b6 — Exception messages passed to BeforeSend now clone ExceptionList, nested stacktrace frames/mechanism pointer fields, and ExceptionFingerprint. Added regression coverage that hook mutations do not alter caller-owned exception data.

posthog.go
marandaneto: Fixed in b1b16b6 — Capture messages passed to BeforeSend now clear Flags and SendFeatureFlags after feature flag properties have already been expanded, with regression coverage that the hook sees both fields as nil while retaining $feature properties.

marandaneto:
we'll have to do a deepcopy then instead of shallow
i dont think we do this for other SDKs tho, so maybe this is overkill?
if anything is wrong we drop the event anyway (panic, invalid type, etc)
the only reason is if the user uses the type concurrently in the beforesend callback after returning, which is misusing the API

ioannisj (approved):
LG. Stamp to unblock but some CI tests failing currently

PR #231, opened 2026-06-19
refactor: organize source and test files

marandaneto:

Keep the repository root less cluttered by organizing library source files under `src/` and root-level tests under `tests/`.

This preserves the existing Go module import path by keeping root-level symlinks to the moved files, and keeps same-package tests working without changing test package visibility.

- `go test ./...`
- `go build .`
- `go vet ./...`
- `make api-diff`
- `go fmt ./...`
- `go mod tidy`
- `git diff --check`
- `go test -v -race -count=1 -timeout=5m ./...`

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

This PR was prepared with the pi coding agent. The main implementation decision was to move physical source and test files into `src/` and `tests/`, while keeping root-level symlinks so `github.com/posthog/posthog-go` remains the module package and existing same-package tests continue to access unexported internals.

The release workflow was updated to sync `src/version.go`, and contributor/release docs were updated to describe the new layout.


marandaneto:
oh that wont work because we will need to keep a symlink for all files

PR #232, opened 2026-06-19
fix: generate UUID v7 by default

marandaneto:

SDK-generated IDs should use UUID v7 so they are time-ordered. This updates the default UUID path used for event `uuid` values and personless distinct IDs created from request context, while preserving caller-provided UUIDs unchanged.

- `go test ./...`

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Implemented by a delegated coding agent from explicit maintainer direction to migrate SDK-generated UUIDs from v4/random to v7. Chose the existing `github.com/google/uuid` dependency's `NewV7()` API and kept explicit user-provided UUID behavior unchanged.


message.go
```diff
@@ -65,12 +65,12 @@ func makeTimestamp(t time.Time, def time.Time) time.Time {
...
 
 // makeUUID returns the UUID passed as first argument if non-empty,
-// otherwise generates and returns a new random UUID (v4).
+// otherwise generates and returns a new UUID v7.
 func makeUUID(u string) string {
 	if u != "" {
 		return u
 	}
-	return uuid.New().String()
+	return uuid.Must(uuid.NewV7()).String()
```
marandaneto, line 73: is that a new Go API? is it supported in the version that this SDK runs on the minimum?

request_context_test.go
```diff
@@ -99,7 +99,9 @@ func TestEnqueueWithContext_MissingIdentityWithRequestContextCreatesPersonlessCa
 	event := readSingleBatchEvent(t, body)
 	distinctID, ok := event["distinct_id"].(string)
 	require.True(t, ok)
-	require.NoError(t, uuid.Validate(distinctID))
+	parsedDistinctID, err := uuid.Parse(distinctID)
+	require.NoError(t, err)
+	require.Equal(t, uuid.Version(7), parsedDistinctID.Version())
```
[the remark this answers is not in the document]
marandaneto: Good catch — fixed in 1ea1906 by adding a shared `requireUUIDv7` test helper and using it for generated personless IDs.

message.go
marandaneto: This comes from `github.com/google/uuid` v1.6.0, not a new Go stdlib API. The SDK minimum is Go 1.21, and `uuid.NewV7()` is available in the existing dependency version, so it is compatible with the supported Go range.

PR #233, opened 2026-06-19
fix: validate supplied event UUIDs

marandaneto:

User-supplied event UUIDs were accepted as-is, so callers could send non-UUID strings in the `uuid` field. This validates supplied UUIDs before enqueueing and falls back to the SDK-generated UUID when a supplied value is invalid.

```bash
go test ./...
```

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Implemented with the pi coding agent. The change keeps valid caller-provided UUIDs intact and replaces invalid UUID strings with generated UUIDs, with regression coverage for valid and invalid supplied UUIDs.


marandaneto:
for all structs we declare the `Uuid string`, lets add a note that the uuid has to be a valid UUID

posthog_test.go
```diff
@@ -746,6 +746,78 @@ func TestEnqueue_UUID_PreSupplied(t *testing.T) {
 	}
 }
 
+// TestEnqueue_UUID_InvalidPreSupplied verifies that invalid user-provided UUIDs
+// are replaced with generated UUIDs instead of being sent as-is.
+func TestEnqueue_UUID_InvalidPreSupplied(t *testing.T) {
+	uuidPattern := regexp.MustCompile(`[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}`)
```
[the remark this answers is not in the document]
marandaneto: Fixed in 601549a — moved the duplicated UUID regex into a shared package-level test variable.

marandaneto:
Addressed in 601549a — added UUID validity notes to every public `Uuid string` field comment and clarified generated wire UUID comments.

posthog_test.go
```diff
@@ -24,6 +24,8 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+var uuidPattern = regexp.MustCompile(`[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}`)
```
marandaneto: this should allow UUIDs from all versions, is that the case?

marandaneto: Yes — the implementation uses `uuid.Validate`, which accepts valid UUID strings without restricting to a specific version. Updated in 7cffaec to use `uuid.Validate` in the tests too, and the preserved-UUID cases now cover multiple UUID versions including v1, v3, v4, v5, and v7.

ioannisj (approved):

PR #234, opened 2026-06-24
fix: Dedupe feature flag called events by response

marandaneto:
`$feature_flag_called` events were deduped only by user, flag, device, and group context. If the same user saw the same flag with different response values in one client process, later exposures could be suppressed incorrectly.

This change includes the feature flag response value in the dedupe cache key so exposures are tracked separately per returned value, while preserving existing device and group-context dedupe behavior.

- `go test ./...`
- `pnpm changeset status --since=origin/main` (failed locally because `node_modules` is not installed; changeset file is included)

## :pencil: Checklist

### If releasing new changes

- [x] Changeset file included for this patch fix.

## 🤖 Agent context

An agent helped prepare this human-directed PR from the existing `fix/ff-called-value-dedupe` worktree. The agent inspected the diff, added the required patch changeset, ran Go tests, committed, pushed, and opened this draft PR for human review.

### DIFF
--- a/feature_flag_evaluations_test.go
+++ b/feature_flag_evaluations_test.go
@@ -743,6 +743,7 @@ func TestCaptureFlagCalled_DedupesByFlagValue(t *testing.T) {
 				cli.captureFlagCalledIfNeeded(
 					"user-1",
 					"changing-flag",
+					tc.response,
 					nil,
 					NewProperties().Set("$feature_flag", "changing-flag").Set("$feature_flag_response", tc.response),
 					nil,
--- a/feature_flag_evaluations_test.go
+++ b/feature_flag_evaluations_test.go
@@ -720,6 +720,52 @@ func TestRefactor_LegacyAndSnapshotPathsDedupeIdentically(t *testing.T) {
 	}
 }
 
+func TestCaptureFlagCalled_DedupesByFlagValue(t *testing.T) {
+	t.Parallel()
+	fs := newFlagsServer(t, "test-flags-v4.json")
+
+	clientIface, capture, _ := newEvalClient(t, fs.server)
+	cli := clientIface.(*client)
+
+	cli.captureFlagCalledIfNeeded(
+		"user-1",
+		"changing-flag",
+		nil,
+		NewProperties().Set("$feature_flag", "changing-flag").Set("$feature_flag_response", true),
+		nil,
+	)
+	cli.captureFlagCalledIfNeeded(
+		"user-1",
+		"changing-flag",
+		nil,
+		NewProperties().Set("$feature_flag", "changing-flag").Set("$feature_flag_response", false),
+		nil,
+	)
+	cli.captureFlagCalledIfNeeded(
+		"user-1",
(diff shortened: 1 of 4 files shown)


posthog.go
```diff
@@ -957,6 +958,7 @@ func (c *client) captureFlagCalledIfNeededWithContext(ctx context.Context, disti
 	cacheKey := flagUser{
 		distinctID: distinctId,
 		flagKey:    key,
+		flagValue:  featureFlagResponseCacheKey(properties["$feature_flag_response"]),
```
marandaneto: receive the `properties["$feature_flag_response"]` as a method param like the others instead of reading from `properties`

feature_flag_evaluations_test.go
```diff
@@ -720,6 +720,52 @@ func TestRefactor_LegacyAndSnapshotPathsDedupeIdentically(t *testing.T) {
...
+	events := snapshotCapturedEvents(capture)
+
+	count := countFlagCalledEvents(events, "changing-flag")
+	if count != 2 {
+		t.Fatalf("expected exactly 2 $feature_flag_called events (true and false), got %d", count)
+	}
+}
+
 func TestCaptureFlagCalled_FiresPerGroupContext(t *testing.T) {
 	t.Parallel()
```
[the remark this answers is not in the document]
marandaneto: Addressed in 6c25b1d: converted the test to table-driven cases and added string and nil response values while still checking duplicate calls only emit once per response.

posthog.go
```diff
@@ -1347,6 +1353,8 @@ func (c *client) CloseWithContext(ctx context.Context) error {
 			// Wait for shutdown to acknowledge cancellation
 			<-c.shutdown
 		}
+
+		c.distinctIdsFeatureFlagsReported.Purge()
```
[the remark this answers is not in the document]
marandaneto: Addressed in 6c25b1d: removed the unrelated cache Purge() from CloseWithContext.

marandaneto:
Addressed Greptile's outside-diff doc-comment feedback in 6c25b1d: featureFlagResponseCacheKey now has its own doc comment and canonicalGroupsRepr's comment is correctly attached.

posthog.go
marandaneto: Addressed in 42f4b95: `captureFlagCalledIfNeeded` now receives the feature flag response as an explicit parameter and uses that for the dedupe cache key instead of reading it back from `properties`.

posthog.go
```diff
@@ -973,6 +975,12 @@ func (c *client) captureFlagCalledIfNeededWithContext(ctx context.Context, disti
 	}
 }
 
+// featureFlagResponseCacheKey returns a stable type-qualified representation
+// of a feature flag response value for exposure deduplication.
+func featureFlagResponseCacheKey(value interface{}) string {
+	return fmt.Sprintf("%T:%v", value, value)
```
dustinbyrne, line 981: good catch on including the type

dustinbyrne (approved):

PR #236, opened 2026-06-25
feat(capture): add v1 send/retry engine (unwired)

eli-r-ph:

Second PR in the opt-in capture-v1 stack (stacked on #235). Adds the **v1 send/retry engine** as new unexported methods on `*client`, unit-tested but **not yet wired** into the live send loop — `processBatch` still calls the legacy `send`. The legacy `/batch/` path is untouched and `api/public-api.txt` is unchanged.

What's here:
- `sendV1` partial-retry state machine: a stable `PostHog-Request-Id` and `created_at` across attempts, an incrementing `PostHog-Attempt`, Bearer auth, `PostHog-Sdk-Info`, and a fresh per-attempt `PostHog-Request-Timestamp`
- partition by per-event result: `ok`/`warning`/unrecognized map to `Success`, `drop` to `Failure`, `retry` is re-sent (only the retry subset on the next attempt), and a uuid absent from `results` is silently dropped
- `isRetryableStatusV1` = {408, 500, 502, 503, 504}; 429 and other 4xx are terminal (matches posthog-rs)
- `Retry-After` honored when it exceeds the configured backoff; shutdown handled via `c.quit`/`c.ctx.Done()` during backoff; max-attempts exhaustion fails the remaining events
- `uploadV1` reuses `c.http` (already carries `BatchUploadTimeout`) and `c.ctx` (no double timeout); `reportV1` treats a malformed 2xx body as terminal for that batch
- `preparedBatch` gains a `uuids` field (nil/unused on the legacy path; populated by the v1 capturer in the next PR)

Per-event and request errors use plain `fmt.Errorf` here; PR5 swaps in the typed `CaptureEventError`/`CaptureRequestError`.

New `capture_v1_send_test.go` against an `httptest.Server`: header presence + stable request-id + incrementing attempt across retries, Bearer token, gzip `Content-Encoding` with a decodable body, all-ok single request, partial retry re-sending only the `retry` subset, terminal results not retried, uuid-missing silent drop, retryable-then-success, parameterized status classification, max-attempts exhaustion, malformed-200 terminal, and `Retry-After` honored. Full `make test` (vet + `-race`) green; `make api-diff` reports no public API change.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Authored with Cursor (Claude Opus 4.8); stacked on #235. The retry/partition semantics mirror posthog-rs (`transport.rs` + `v1_capture.rs`): only `retry` is non-terminal, the request-id is stable for a batch's lifetime, and 429 is intentionally terminal in v1 unlike legacy. Errors are intentionally untyped in this PR to keep it focused; the typed-error swap is a later PR so reviewers can see the engine and the public error surface separately.

### DIFF
--- a/capture_v1_send_test.go
+++ b/capture_v1_send_test.go
@@ -519,6 +519,7 @@ func TestV1SendRetryAfterHonored(t *testing.T) {
 	if err1 != nil || err2 != nil {
 		t.Fatalf("parse timestamps: %v / %v", err1, err2)
 	}
+	// RFC3339 has 1-second resolution; Retry-After: 1 means ≥900ms is expected.
 	delta := t2.Sub(t1)
 	if delta < 900*time.Millisecond {
 		t.Errorf("attempt delta %v, expected >= ~1s from Retry-After", delta)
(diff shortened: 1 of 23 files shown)


capture_v1_send_test.go
```diff
@@ -0,0 +1,626 @@
...
+
+	c := newV1TestClient(t, ts.URL, cb, 9, nil)
+	c.sendV1(v1Batch(t, cap1(uuidA)))
+
+	if len(srv.snapshot()) != 1 {
+		t.Fatalf("expected 1 request, got %d", len(srv.snapshot()))
+	}
+	if s, f := cb.counts(); s != 0 || f != 0 {
+		t.Errorf("callbacks: success=%d failure=%d, want 0/0 (silent drop)", s, f)
+	}
```
dustinbyrne, line 398: is this an expected backend response? where the UUID is not returned?

dustinbyrne (approved):
question about the case where a UUID is not returned, but otherwise this change looks good

capture_v1_send_test.go
eli-r-ph: The backend will always return all UUIDs, this is just defensive, same is done in rust SDK 👍

turnipdabeets (approved):

PR #237, opened 2026-06-25
feat(capture): route capture between legacy and v1 via CaptureMode

eli-r-ph:

Third PR in the opt-in capture-v1 stack (stacked on #236). This is the **wire-in**: it makes the v1 serialization (#235) and send engine (#236) reachable from the live path, gated by a new `Config.CaptureMode` that **defaults to legacy**. Merging the stack is transparent to existing callers and ships as a **minor** release; the major bump / default flip / legacy removal remain a separate future workstream.

What's here:
- `CaptureMode` enum (`CaptureModeLegacy` = 0 default, `CaptureModeAnalyticsV1` = 1) + `Config.CaptureMode`, validated in `Config.Validate`
- a `capturer` strategy interface with `legacyCapturer` (the existing `/batch/` serialize + send) and `analyticsV1Capturer` (the PR1 serialization + PR2 send engine); the client picks one in `NewWithConfig` from `CaptureMode`
- `Enqueue` now serializes via `c.capture.prepare` and threads the per-event uuid through `preparedMessage`/`preparedBatch` and the batching loop; `processBatch` dispatches via `c.capture.send`
- `MockServerBuilder` gains a capture-v1 route + request-path recording for tests

### Compression codecs (zstd / deflate / brotli)

The v1 endpoint decodes four `Content-Encoding` codecs; previously the SDK only emitted gzip. This PR adds the other three:
- `CompressionZstd` (2), `CompressionDeflate` (3), `CompressionBrotli` (4) on the existing `CompressionMode`, plus a `String()` helper for log/error readability (config.go).
- a `compressV1Body` helper in capture_v1_send.go that emits the matching wire token (`gzip` / `deflate` / `zstd` / `br`); `uploadV1` sets `Content-Encoding` from it.
- **Deflate is emitted as zlib (RFC 1950)** so the stream starts with `0x78`; the backend sniffs that first byte to route it to its zlib decoder deterministically (server-side first-byte sniffing landed in posthog/posthog#62827), sidestepping raw-vs-zlib ambiguity.
- **Gating:** the three new codecs require `CaptureModeAnalyticsV1`. The legacy `/batch/` endpoint only understands gzip/lz64/base64, so `Config.Validate` rejects zstd/deflate/brotli under `CaptureModeLegacy`. gzip and none stay valid on both modes.

**Dependencies (pure Go, no cgo):** stdlib `compress/zlib` (deflate); `github.com/klauspost/compress/zstd` v1.17.11 (zstd, shared concurrency-safe encoder via `EncodeAll`); `github.com/andybalholm/brotli` v1.1.1 (brotli). klauspost is pinned to v1.17.x deliberately to keep the module `go 1.21` floor (v1.18+ needs Go 1.22). `CGO_ENABLED=0 go build` verified.

The public API gains only additive symbols (`CaptureMode`, the three `Compression*` constants, `CompressionMode.String`); two `minor` changesets are included.

New `capture_v1_integration_test.go`: mode selection (default `Config{}` hits `/batch/`, `CaptureModeAnalyticsV1` hits `/i/v1/analytics/events`), the v1 envelope + `Authorization: Bearer` / `PostHog-Sdk-Info` headers end-to-end through `Enqueue`, and partial retry firing per-event `Success`/`Failure` (3 events: ok / drop / retry to ok = 2 success, 1 failure). Legacy tests are untouched and still assert `/batch/`.

Compression coverage:
- `TestV1SendCompressionCodecs` (table-driven) drives the full send path for none/gzip/zstd/deflate/brotli and asserts the right `Content-Encoding`; the test server decodes per that header and recovers the original batch.
- `TestCompressV1Body` checks the codec helper directly: correct wire token, real size reduction on compressible input, clean round-trip, and the unknown-mode error path.
- `TestConfigCompression` (table-driven) covers each codec x capture mode: valid / v1-only-rejected / unknown-invalid.

Cross-language decode proof: the `capture-v1-go-smoke` and `capture-v1-rs-smoke` rigs compress with the same libraries used here (klauspost zstd, andybalholm brotli, stdlib zlib) and assert `result: ok` against a live v1 backend, including the zlib-wrapped-deflate case.

Full `make test` (vet + `-race`) green; `make api-diff` clean after `make api-update`.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Authored with Cursor (Claude Opus 4.8); stacked on #236. The `capturer` interface is the single encapsulation point for every legacy-vs-v1 divergence (both serialization shape and send/retry semantics), chosen once at construction so the public API, queue, and batching machinery stay protocol-agnostic. Building v1 behind this seam means the eventual major-version cutover is mostly deletion (drop `legacyCapturer`, hardcode v1, remove the enum branch).

The compression codecs were added in this PR (rather than a later one) because it is the earliest layer where `CaptureMode` exists, so the v1-only gating can be enforced atomically with the codecs themselves. #238 and #239 were rebased on top; their content is unchanged (range-diff verified).

### DIFF
--- a/capture_v1_send.go
+++ b/capture_v1_send.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"net/http"
 	"strconv"
+	"sync"
 	"time"
 
 	"github.com/andybalholm/brotli"
@@ -17,16 +18,14 @@ import (
 	"github.com/klauspost/compress/zstd"
 )
 
-// zstdEncoder is shared across all clients: klauspost's EncodeAll is safe for
-// concurrent use and the library recommends reusing one encoder over
-// allocating per call. The options are static so NewWriter cannot fail here.
-var zstdEncoder = func() *zstd.Encoder {
-	enc, err := zstd.NewWriter(nil, zstd.WithEncoderLevel(zstd.SpeedDefault))
-	if err != nil {
-		panic(fmt.Sprintf("posthog: zstd encoder init: %v", err))
-	}
-	return enc
-}()
+// getZstdEncoder returns a shared zstd encoder, lazily initialized on first
+// call. klauspost's EncodeAll is safe for concurrent use and the library
+// recommends reusing one encoder over allocating per call. The init is
+// practically infallible with static options, but returning an error keeps
+// SDK setup graceful (no panic on load).
+var getZstdEncoder = sync.OnceValues(func() (*zstd.Encoder, error) {
+	return zstd.NewWriter(nil, zstd.WithEncoderLevel(zstd.SpeedDefault))
+})
 
 // v1Result is the parsed outcome of a single capture-v1 HTTP attempt.
 type v1Result struct {
@@ -239,7 +238,11 @@ func compressV1Body(mode CompressionMode, raw []byte) ([]byte, string, error) {
 		}
 		return buf.Bytes(), "deflate", nil
 	case CompressionZstd:
-		return zstdEncoder.EncodeAll(raw, make([]byte, 0, len(raw))), "zstd", nil
+		enc, err := getZstdEncoder()
+		if err != nil {
+			return nil, "", fmt.Errorf("zstd encoder init: %w", err)
+		}
+		return enc.EncodeAll(raw, make([]byte, 0, len(raw))), "zstd", nil
 	case CompressionBrotli:
 		var buf bytes.Buffer
 		bw := brotli.NewWriter(&buf)
--- a/capture_v1_send.go
+++ b/capture_v1_send.go
@@ -3,6 +3,7 @@ package posthog
 import (
 	"bytes"
 	"compress/gzip"
+	"compress/zlib"
 	"context"
 	"fmt"
 	"io"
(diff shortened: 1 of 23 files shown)


capture_v1_send.go
```diff
@@ -3,17 +3,31 @@ package posthog
...
 )
 
+// zstdEncoder is shared across all clients: klauspost's EncodeAll is safe for
+// concurrent use and the library recommends reusing one encoder over
+// allocating per call. The options are static so NewWriter cannot fail here.
+var zstdEncoder = func() *zstd.Encoder {
+	enc, err := zstd.NewWriter(nil, zstd.WithEncoderLevel(zstd.SpeedDefault))
+	if err != nil {
+		panic(fmt.Sprintf("posthog: zstd encoder init: %v", err))
+	}
```
dustinbyrne: we're sure this can't return an error given these static options in _any_ environment?

panicking on init is quite scary - can we avoid it?

eli-r-ph: Oops, updated some of the encoders as a last pass and that must have snuck in - I agree that's not ideal, will update thanks! 🙇

eli-r-ph: fixed, thanks for the callout!

turnipdabeets (approved):

PR #249, opened 2026-07-01
ci: skip heavy CI for non-impacting changes

marandaneto:
## Summary
- update the existing Unit Tests and Formatting skip logic to use a generic safe-file classifier instead of SDK/repo-specific source/test/build allowlists
- only skip on `pull_request` when every changed `filename` and non-empty `previous_filename` from the GitHub PR files API is globally safe
- keep required check/job names materialized while no-oping checkout/setup-go/Go commands for safe changes; fail open to full checks for pushes, merge groups, API/payload failures, empty file lists, current CI workflow edits, source/test/build/package/lock/config files, and unknown paths

## Validation
- Ruby YAML parse for `.github/workflows/unit-tests.yml` and `.github/workflows/fmt.yml`
- `bash -n` on extracted classifier scripts from both workflows
- mocked classifier sample cases for both workflows: generic safe set => `run_go_checks=false`; source file, unsafe previous filename, current CI workflow, package/lock files, empty API list, push event, and API failure => `run_go_checks=true`
- `git diff --check`

### DIFF
--- a/.github/workflows/fmt.yml
+++ b/.github/workflows/fmt.yml
@@ -15,36 +15,174 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  changes:
+    name: Classify changes
+    runs-on: ubuntu-latest
+    outputs:
+      run_go_checks: ${{ steps.classify.outputs.run_go_checks }}
+      reason: ${{ steps.classify.outputs.reason }}
+    steps:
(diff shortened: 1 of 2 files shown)


marandaneto:
Closing this for now — I'll investigate other options. We don't want hardcoded paths this way.

PR #253, opened 2026-07-02
feat: add SecretKey config, deprecate PersonalApiKey

turnipdabeets:
## Problem

The config field used for local feature flag evaluation and remote config is named `PersonalApiKey`, but it actually accepts **either** a Personal API Key (`phx_...`) **or** a Project Secret API Key (`phs_...`). The name is misleading — users with a Project Secret API Key reasonably assume the field won't accept their key.

## Change

- Add a new canonical config field **`SecretKey`** that accepts either credential type.
- Keep **`PersonalApiKey`** as a **deprecated alias** (marked `// Deprecated: use SecretKey instead.`). Fully backwards compatible / non-breaking.
- Resolve the effective credential at the point of use: prefer `SecretKey`, fall back to `PersonalApiKey`. `SecretKey` wins when both are set.
- Update user-facing godoc and the "not configured" log message to reference `SecretKey`.
- Tests cover: `SecretKey` is used, `PersonalApiKey` still works, and `SecretKey` wins when both are set (unit resolution + end-to-end remote-config `Authorization` header).

`go build ./...`, `go vet ./...`, and `go test ./...` all pass.

## Context

Coordinated rename across the backend SDKs. See internal Slack discussion and PostHog/posthog-js#4046 (the field-rename item only). Mirrors the draft opened in posthog-python (#727).

### DIFF
--- a/posthog_test.go
+++ b/posthog_test.go
@@ -275,6 +275,40 @@ func TestNewWithConfig_TrimsWhitespaceSensitiveInputsInRequests(t *testing.T) {
 	require.Equal(t, "Bearer test-personal-key", remoteConfigAuth)
 }
 
+func TestRemoteConfigAuthUsesResolvedSecretKey(t *testing.T) {
+	cases := []struct {
+		name     string
+		config   Config
+		wantAuth string
+	}{
+		{"secret key", Config{SecretKey: "phs_secret"}, "Bearer phs_secret"},
+		{"personal api key fallback", Config{PersonalApiKey: "phx_personal"}, "Bearer phx_personal"},
+		{"secret key wins", Config{SecretKey: "phs_secret", PersonalApiKey: "phx_personal"}, "Bearer phs_secret"},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			var gotAuth string
+			server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				gotAuth = r.Header.Get("Authorization")
+				w.Header().Set("Content-Type", "application/json")
+				_, _ = w.Write([]byte(`"ok"`))
+			}))
+			defer server.Close()
+
+			tc.config.Endpoint = server.URL
+			client, err := NewWithConfig("test-api-key", tc.config)
+			require.NoError(t, err)
+			defer client.Close()
+
+			payload, err := client.GetRemoteConfigPayload("test-flag")
+			require.NoError(t, err)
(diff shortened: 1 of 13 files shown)


turnipdabeets:
**Marked ready for review.**

Naming is settled — this is a generic **`secret_key`** config that accepts **either** a Personal API Key or a Project Secret API Key (both coexist; same `Authorization: Bearer` header). `personal_api_key` stays as a deprecated **config-name** alias that forwards into it — *not* a deprecation of personal keys as a credential. It's a name-level change and is crash-safe with respect to using the key (adds no new error path).

Merge timing still coordinates with the project-secret-key rollout (feature-flagged today) — ready-for-review ≠ merge-now.

posthog_test.go
```diff
@@ -275,6 +275,47 @@ func TestNewWithConfig_TrimsWhitespaceSensitiveInputsInRequests(t *testing.T) {
...
+				_, _ = w.Write([]byte(`"ok"`))
+			}))
+			defer server.Close()
+
+			tc.config.Endpoint = server.URL
+			client, err := NewWithConfig("test-api-key", tc.config)
+			require.NoError(t, err)
+			defer client.Close()
+
+			payload, err := client.GetRemoteConfigPayload("test-flag")
```
dustinbyrne: non-blocking, but non-obvious: `NewWithConfig` can launch the flags poller which will load feature flag definitions with the auth header

not sure if `GetRemoteConfigPayload` is even needed?

dustinbyrne (approved):
some references remain:
```sh
$ git -P grep -l 'personal_api_key' -- . ':(exclude)references/**'
examples/.env.example
examples/README.md
feature_flags_local_test.go
```

posthog_test.go
turnipdabeets: Good callouts. The poller-launches-with-auth-header point is exactly why this test guards the captured header with a mutex — both the poller and the remote-config request hit the httptest server, which was a real data race (fixed here). On whether `GetRemoteConfigPayload` is needed: I kept it as the assertion target since it exercises the resolved `SecretKey` on the Authorization header, but happy to switch to a poller-based assertion if you prefer. Non-blocking, leaving as-is.

turnipdabeets: Followed up on this — reworked the test to skip the poller entirely. It now constructs a minimal `client` and calls `makeRemoteConfigRequest` directly (no `NewWithConfig`, so no background flags poller fires). That removed the data race and the mutex; the httptest handler runs once, synchronously. `GetRemoteConfigPayload` is no longer involved in the test. (d91f2e9)

PR #255, opened 2026-07-03
fix(capture): unify retry backoff ceiling at 30s

eli-r-ph:

Cross-SDK capture retry parity. This unifies the retry backoff ceiling at a 30s default across posthog-go, posthog-rs, and posthog-python, and bounds a server `Retry-After` (previously unbounded in Go) so a hostile/buggy header can't stall a batch goroutine.

- `retryDelayV1` now waits the larger of the configured backoff and the `Retry-After`, with the header clamped to `defaultMaxBackoff` (30s); the configured backoff is never truncated.
- `DefaultBackoff` cap changed 10s -> 30s (default only; override via `Config.RetryAfter`), unifying Go's exp-backoff ceiling with the other SDKs.
- `go build`, `gofmt`, and `go test ./...` all green; `TestRetryDelayV1` updated for the clamp (absurd/above-ceiling -> 30s).

## :pencil: Checklist

- [x] Default behavior change (`DefaultBackoff` 10s -> 30s; `Retry-After` clamped) — changeset added (`minor`).

### If releasing new changes

- [x] Added a changeset

## 🤖 Agent context

Authored with Cursor (Claude Opus 4.8) as part of a cross-SDK capture retry unification (paired with posthog-rs #177 and the posthog-python capture-v1 stack). Prod telemetry showed the backend sends `Retry-After: 1s` and retryable responses are ~1 in 18M, so 30s (the existing default ceiling elsewhere) is the agreed unified value.

### DIFF
--- a/.changeset/capture-v1-retry-after-cap.md
+++ b/.changeset/capture-v1-retry-after-cap.md
@@ -1,0 +1,5 @@
+---
+"posthog-go": minor
+---
+
+Unify the capture retry backoff ceiling at 30s. `DefaultBackoff`'s cap changes from 10s to 30s (default only — override via `Config.RetryAfter`), and the Capture V1 send now clamps a server `Retry-After` to the same 30s so a hostile or buggy header cannot park a batch goroutine. `Retry-After` still acts as a minimum; the configured backoff is never truncated. This aligns the default retry behavior with posthog-rs and posthog-python.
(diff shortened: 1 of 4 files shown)


.changeset/capture-v1-retry-after-cap.md
```diff
@@ -0,0 +1,5 @@
+---
+"posthog-go": minor
```
marandaneto: i'd say its a patch

marandaneto (approved):

PR #256, opened 2026-07-03
feat(capture): default to 4 delivery attempts (down from 10)

eli-r-ph:

Cross-SDK capture parity. posthog-go defaults to 10 delivery attempts when `Config.MaxRetries` is unset, while posthog-rs/posthog-python default to 4. This changes the Go default to 4 (`DefaultMaxAttempts`) to match.

Note: `maxAttempts` is shared by the v0 and v1 send paths, so this also lowers v0's default retry count. Callers that set `MaxRetries` explicitly are unaffected. Draft pending maintainer sign-off on the default change.

- `Validate()` uses `DefaultMaxAttempts` (4) when `MaxRetries` is unset or out of range.
- `go build`, `gofmt`, and `go test ./...` all green; config/retry/tcp-drop tests that were calibrated to 10 updated to `DefaultMaxAttempts`.

## :pencil: Checklist

- [x] Default behavior change (attempt budget 10 -> 4) — changeset added (`minor`).

### If releasing new changes

- [x] Added a changeset

## 🤖 Agent context

Authored with Cursor (Claude Opus 4.8) during a cross-SDK capture parity pass. Left as a draft because the shared `maxAttempts` means this touches v0 default retries too — flagged for maintainer sign-off on scope (shared vs v1-only).

### DIFF
--- a/.changeset/capture-default-max-attempts.md
+++ b/.changeset/capture-default-max-attempts.md
@@ -1,0 +1,5 @@
+---
+"posthog-go": minor
+---
+
+Change the default capture delivery budget from 10 attempts to 4 (`DefaultMaxAttempts`) when `Config.MaxRetries` is unset, aligning with the cross-SDK Capture V1 parity standard (posthog-rs uses the same envelope). This affects **both** the v0 (`/batch/`) and v1 send paths, since they share the attempt budget. Callers that set `MaxRetries` explicitly are unaffected.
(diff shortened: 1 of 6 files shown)


.changeset/capture-default-max-attempts.md
```diff
@@ -0,0 +1,5 @@
+---
+"posthog-go": minor
```
marandaneto: should it be a patch?

marandaneto (approved):

PR #264, opened 2026-07-31
feat: support starts_with and ends_with operators in local evaluation

haacked:

PostHog/posthog#72992 added four case-insensitive property filter operators — `starts_with`, `not_starts_with`, `ends_with`, `not_ends_with` — across HogQL, the flags service, and the UI. The UI surface is gated behind a feature flag until server-side SDKs can evaluate these operators locally; on SDK versions without support, flags using them fall back to remote evaluation.

This adds local evaluation support for the four operators, mirroring `icontains`: both sides are stringified and lowercased, and the `not_` variants are exact negations. Filter values are single strings — server-side validation rejects list values for these operators, matching the flags service behavior.

New unit tests mirror the flags service's own test matrix for these operators: case-insensitive anchored matching, mid-string non-matches (`prevalue` does not match `starts_with: "Val"`), numeric stringification (`323` matches `starts_with: "3"`, `123` does not), negation inverses, and missing-key inconclusive behavior.

`go test -run TestMatchProperty .` passes locally (including the new `TestMatchPropertyStartsWith` and `TestMatchPropertyEndsWith`); `gofmt` and `go vet` are clean.

## :pencil: Checklist

### If releasing new changes

- [x] Added a changeset file (minor)

---

### DIFF
--- a/feature_flags_matching_test.go
+++ b/feature_flags_matching_test.go
@@ -351,9 +351,26 @@ func TestMatchPropertyStartsWith(t *testing.T) {
 		require.True(t, isMatch, "expected %v to match", val)
 	}
 
-	_, err = matchProperty(property, NewProperties().Set("key2", "value"))
+	// The negated operator is the exact inverse across numeric values too.
+	negatedNumericProperty := FlagProperty{
+		Key:      "key",
+		Value:    "3",
+		Operator: "not_starts_with",
+	}
+
+	isMatch, err = matchProperty(negatedNumericProperty, NewProperties().Set("key", 323))
+	require.NoError(t, err)
+	require.False(t, isMatch)
+
+	isMatch, err = matchProperty(negatedNumericProperty, NewProperties().Set("key", 123))
+	require.NoError(t, err)
+	require.True(t, isMatch)
+
 	var inconclusiveErr *InconclusiveMatchError
-	require.ErrorAs(t, err, &inconclusiveErr)
+	for _, missing := range []Properties{NewProperties().Set("key2", "value"), NewProperties()} {
+		_, err = matchProperty(property, missing)
+		require.ErrorAs(t, err, &inconclusiveErr)
+	}
 }
 
 func TestMatchPropertyEndsWith(t *testing.T) {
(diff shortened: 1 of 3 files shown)


marandaneto (approved):

feature_flags_matching_test.go
```diff
@@ -299,6 +299,124 @@ func TestMatchPropertyContains(t *testing.T) {
...
+	require.NoError(t, err)
+	require.False(t, isMatch)
+
+	negatedProperty := FlagProperty{
+		Key:      "key",
+		Value:    "Val",
+		Operator: "not_starts_with",
+	}
+
+	for _, val := range []interface{}{"value", "VALUE"} {
```
marandaneto, line 342: Non-blocking test gap: the positive operators exercise numeric stringification, but the negated cases only use strings. Could we add numeric cases for `not_starts_with` and `not_ends_with` (or assert that each negated operator is the exact inverse across the positive operator’s table)? The implementation uses direct `!`, so this is primarily regression coverage.

feature_flags_matching_test.go
```diff
@@ -299,6 +299,124 @@ func TestMatchPropertyContains(t *testing.T) {
...
+		require.False(t, isMatch, "expected %v not to match", val)
+	}
+
+	for _, val := range []interface{}{"prevalue", "Alakazam"} {
+		isMatch, err := matchProperty(negatedProperty, NewProperties().Set("key", val))
+		require.NoError(t, err)
+		require.True(t, isMatch, "expected %v to match", val)
+	}
+
+	_, err = matchProperty(property, NewProperties().Set("key2", "value"))
```
marandaneto: Small coverage suggestion: Python’s parity table checks both an empty property map and a map containing an unrelated key. Could we also exercise `NewProperties()` here (and for the suffix operators)? The shared missing-key guard already looks correct; this would just cover both input forms explicitly.

featureflags.go
```diff
@@ -1376,6 +1376,22 @@ func matchProperty(property FlagProperty, properties Properties) (bool, error) {
 		return !strings.Contains(strings.ToLower(valueToString(override_value)), strings.ToLower(valueToString(value))), nil
 	}
 
+	if operator == "starts_with" {
+		return strings.HasPrefix(strings.ToLower(valueToString(override_value)), strings.ToLower(valueToString(value))), nil
```
marandaneto, line 1380: Non-blocking parity note: Go’s `strings.ToLower` is not equivalent to Python’s Unicode `casefold`, so rare values such as `ß` versus `SS` can differ. This is consistent with the existing Go `icontains` implementation (and the server uses ASCII lowering), so it may be intentional; a test or brief documentation of the expected Unicode behavior would make that contract explicit.

featureflags.go
```diff
@@ -1376,6 +1376,22 @@ func matchProperty(property FlagProperty, properties Properties) (bool, error) {
...
+	}
+
+	if operator == "not_starts_with" {
+		return !strings.HasPrefix(strings.ToLower(valueToString(override_value)), strings.ToLower(valueToString(value))), nil
```
marandaneto, line 1384: Non-blocking SDK-specific edge case: `valueToString(nil)` produces `"<nil>"`, so a negated operator can return `true` for `nil`, whereas Python returns `false` for all four operators when the override is `None`. Python only requires exact inversion for non-`None` values, and this follows Go’s existing stringification behavior; a focused `nil` test would lock in the intended Go contract.

marandaneto (commented):
Non-blocking notes from the cross-SDK parity review.

featureflags.go
haacked: Fixed!

featureflags.go
haacked: Fixed!

Issue #262, opened 2026-07-20
Are you planning to add official MCP SDK support for Go?


wolfeidau:
### Problem Statement

Keen to understand if this is planned or your open to having it contributed.

Cheers

pauldambra:
Hey 👋 we're absolutely planning to extend coverage. And ideally based on folks asking us :)

If you're open to contributing then you can see how we wrap the Node SDK as an example https://github.com/PostHog/posthog-js/tree/main/packages/mcp

That's what we'd copy if we were shipping it

(cc @lucasheriques and @gesh )

wolfeidau:
@pauldambra I am just going through https://github.com/PostHog/posthog-js/pull/3653 at the moment, not sure I will start with something this size though. 😅 I may start with something a little smaller.

This PR has lots of good context though.

lucasheriques:
@wolfeidau if you have any questions feel free to ask! it's ok to start small, one PR for the interfaces/structure/types, another PR for maybe one part of it.

also, pls notice [the current MCP is about to change](https://blog.modelcontextprotocol.io/posts/2026-07-28-release-candidate/) and we want to make sure we support it. The biggest change that affects us is making sure we can identify users when MCP is stateless

i'd be happy to review any PRs; feel free to rag me directly

wolfeidau:
@lucasheriques I have put together a working version in #265 on top of this is a wrapper for the Go MCP sdk https://github.com/wolfeidau/posthog-go/commit/d6505e060a26c83661cf3c0832c3c41006750e6e 

There is a spec in #265 which I mostly followed, it outlines some of the tradeoffs with Go versions support, and the use of a seperate Go Module for the actual wrapper. 

Still quite a few things to sort out but most of the plumbing is there.

PR #265, opened 2026-08-02
feat(mcp): add canonical MCP tool-call events

wolfeidau:
Verified with go test ./..., MCP race tests, go vet, and the public API check.

- add a framework-independent MCP analytics API with identity mapping and optional exception fanout
- sanitize and bound captured parameters, responses, properties, and errors
- add API documentation, provenance, fixtures, and tests

I have left the original specification in the mcp_support_spec.md happy to remove it.

Screenshot 2026-08-02 at 11 50 14 am

This is half of the solution to #262

This is been verified on my personal side project https://github.com/wolfeidau/starlogz 

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

This code was initially scaffolded by OpenAI Codex, and most of the test code was written with help from this agent harness.

**Autonomy:** Human-driven (agent-assisted) — or — Fully autonomous

### DIFF
--- a/mcp/event.go
+++ b/mcp/event.go
@@ -1,0 +1,261 @@
+package mcp
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+	"time"
+
+	posthog "github.com/posthog/posthog-go"
+)
+
+type preparedToolCall struct {
+	call          ToolCall
+	distinctID    string
+	explicitID    bool
+	toolName      string
+	intent        string
+	intentSource  IntentSource
+	errorType     string
+	errorMessage  string
+	parameters    any
+	response      any
+	groups        posthog.Groups
+	setProperties posthog.Properties
+	custom        posthog.Properties
+}
+
+func buildToolCallMessages(call ToolCall, exceptionAutocapture bool) ([]namedMessage, error) {
+	prepared, err := prepareToolCall(call)
+	if err != nil {
+		return nil, err
+	}
+
+	capture, err := prepared.buildCapture()
+	if err != nil {
+		return nil, err
+	}
+	messages := []namedMessage{{name: eventToolCall, message: capture}}
+
+	if call.IsError && exceptionAutocapture {
+		exception, err := prepared.buildException()
+		if err != nil {
+			return nil, err
+		}
+		messages = append(messages, namedMessage{name: "$exception", message: exception})
+	}
+	return messages, nil
+}
+
+func prepareToolCall(call ToolCall) (preparedToolCall, error) {
+	if strings.TrimSpace(call.ToolName) == "" {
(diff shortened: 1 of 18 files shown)


marandaneto:
@PostHog/mcp-analytics pls take a look!

mcp/event.go
```diff
@@ -0,0 +1,261 @@
...
+			resultErr = errors.New("posthogmcp: Error method panicked")
+		}
+	}()
+	return err.Error(), nil
+}
+
+func removeIdentityControlProperties(properties posthog.Properties) {
+	delete(properties, propertyGroups)
+	delete(properties, propertySet)
+	delete(properties, propertyProcessProfile)
```
lucasheriques, line 188: Preserve explicit person-profile opt-out

Passing `DistinctID: "user_1"` with `Properties{"$process_person_profile": false}` drops the opt-out from both `$mcp_tool_call` and `$exception`. Python and TypeScript preserve it. I reproduced the missing flag in both serialized payloads.

Please preserve an explicitly supplied `false` for identified calls, while continuing to force `false` when identity is absent. Add serialized-payload tests for both events, including an anonymous call attempting to override the flag to `true`.

mcp/event.go
```diff
@@ -0,0 +1,261 @@
...
+		prepared.distinctID = call.DistinctID
+	case call.SessionID != "":
+		prepared.distinctID = call.SessionID
+	default:
+		prepared.distinctID = "anonymous"
+	}
+
+	intent := strings.TrimSpace(call.Intent)
+	if intent != "" {
+		prepared.intent = truncateUTF8(sanitizeString(intent), maxIntentBytes)
```
lucasheriques, line 79: Bring intent redaction up to current SDK parity

Both Python and TypeScript now redact structured personal identifiers from `$mcp_intent`. This implementation still captures emails and IP addresses unchanged: `Find orders for alice@example.com from 192.0.2.1` is emitted verbatim, while both current SDKs produce `Find orders for [redacted] from [redacted]`. This behavior landed after the PR's August baseline.

Please port the intent-specific redaction for emails, phone numbers, IPv4/IPv6 addresses, Luhn-valid cards, and US SSNs. Run it before truncation and reuse the cross-language test cases, including Unicode separators and false positives. Keep this additional redaction scoped to intent; parameters and responses should retain their existing behavior.

References: [TypeScript sanitizer](https://github.com/PostHog/posthog-js/blob/891eefa06b73e2259845e625769161ae4a762fd7/packages/mcp/src/extensions/mcp-payloads.ts), [Python sanitizer](https://github.com/PostHog/posthog-python/blob/e823b3e557e666fda72516d8ed75630ceb758559/posthog/mcp/_sanitization.py).

turnipdabeets:
@wolfeidau friendly ping

PR #284, opened 2026-08-14
fix: normalize SDK timestamps to UTC

marandaneto:

Event timestamps supplied with a non-UTC location were serialized with their original offset. This made SDK payload timestamps inconsistent even though they represented the correct instant.

This change converts event timestamps to the equivalent UTC instant before serializing legacy batch and Capture V1 payloads. It covers Capture, Identify, GroupIdentify, Alias, and Exception messages. Default timestamps from the client clock are also normalized. Timestamp values stored in caller properties are left unchanged.

- `TestEventTimestampSerializationUsesUTC` verifies explicit non-UTC timestamps for Capture, Identify, GroupIdentify, Alias, and Exception messages in both legacy and Capture V1 payloads.
- `TestTimestampNormalizationDoesNotRewriteCallerProperties` verifies that only the event timestamp is normalized and a time value in caller properties keeps its original offset.
- `TestDefaultEventTimestampsUseUTC` verifies that zero-value event timestamps use the client clock and serialize in UTC for all five message types.
- `TestV1SendHeadersAndAllOk` verifies that `PostHog-Request-Timestamp` and the Capture V1 `created_at` field use the same UTC instant from a non-UTC client clock.
- Ran the focused timestamp tests and `go vet ./...`.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Prepared with Pi's PR workflow and isolated autoreview. The requested behavior and scope were directed by @marandaneto. The local Pi session has no public URL. Autoreview reported no actionable findings.

### DIFF
--- a/.changeset/calm-clocks-align.md
+++ b/.changeset/calm-clocks-align.md
@@ -1,0 +1,5 @@
+---
+"posthog-go": patch
+---
+
+Normalize event timestamps to the equivalent UTC instant before serializing legacy batch and Capture V1 payloads.
--- a/alias.go
+++ b/alias.go
@@ -21,7 +21,8 @@ type Alias struct {
 	Alias string
 	// DistinctId is the existing user distinct ID that Alias should resolve to.
 	DistinctId string
-	// Timestamp is the event timestamp. If zero, Enqueue uses the current time.
+	// Timestamp is the event timestamp. UTC is preferred; non-UTC values are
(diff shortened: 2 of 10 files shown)


arnohillen (approved):
<small>(created by claude code)</small>

[arno's agent] Verified at 7a022301: .UTC() is instant-preserving and idempotent so the makeTimestamp+APIfy double application is harmless, and all five message types normalize at both Enqueue and serialization.

PR #286, opened 2026-08-19
fix(flags): fall back for missing local definitions

marandaneto:

When `EvaluateFlags` received `FlagKeys` containing a key absent from loaded local definitions, the SDK treated local evaluation as complete and omitted that key. It did not ask `/flags`, even though the caller had not requested local-only evaluation.

The server SDK behavior was agreed in [PostHog/posthog-android#678](https://github.com/PostHog/posthog-android/pull/678#issuecomment-5341879578): a requested key missing from local definitions should fall back to `/flags` because callers can use `OnlyEvaluateLocally` when they need a strictly local result. The canonical contract is being updated in [PostHog/sdk-specs#44](https://github.com/PostHog/sdk-specs/pull/44).

### Changes

- Mark explicitly requested keys missing from local definitions as unresolved and use the existing single remote fallback.
- Send the caller's original key list to `/flags`, keep locally resolved values authoritative, and exclude extra response keys from the snapshot.
- Keep `OnlyEvaluateLocally` strictly local. Missing and inconclusive keys remain absent without a remote request.
- Document that `EvaluateFlags` does not cache snapshots. A key absent both locally and remotely can therefore cause one `/flags` request on each call.
- Use a minor changeset because this reverses observable request behavior.

- `go test ./... -run 'TestEvaluateFlags_(MissingRequestedLocalFlagFallsBackRemotely|MissingRequestedFlagAbsentRemotelyRequestsEachCall|ForwardsFlagKeys)$' -count=1`
- `go vet ./...`
- `pnpm exec changeset status`
- `make test`

The regression tests cover missing-key fallback, original request scope, local-over-remote precedence, response scoping, strict local-only behavior, and the per-call cost when neither local definitions nor `/flags` know the key.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Implemented with Pi's worker agent from a human-directed cross-SDK consistency task. The patch follows the behavior agreed in PostHog/posthog-android#678 and the canonical contract proposed in PostHog/sdk-specs#44.

### DIFF
--- a/.changeset/quiet-flags-fallback.md
+++ b/.changeset/quiet-flags-fallback.md
@@ -1,0 +1,5 @@
+---
+"posthog-go": patch
+---
+
+Fall back to `/flags` when a requested flag is missing from loaded local definitions.
--- a/.changeset/quiet-flags-fallback.md
+++ b/.changeset/quiet-flags-fallback.md
@@ -1,5 +1,5 @@
 ---
-"posthog-go": patch
+"posthog-go": minor
 ---
 
-Fall back to `/flags` when a requested flag is missing from loaded local definitions.
+Fall back to `/flags` when a requested flag is missing from loaded local definitions. This changes the earlier behavior where the key was omitted without a request.
--- a/feature_flag_evaluations_test.go
+++ b/feature_flag_evaluations_test.go
@@ -579,6 +579,84 @@ func TestEvaluateFlags_ForwardsFlagKeys(t *testing.T) {
 	}
 }
 
+func TestEvaluateFlags_MissingRequestedLocalFlagFallsBackRemotely(t *testing.T) {
+	t.Parallel()
+	var remoteCalls atomic.Int32
+	remoteRequests := make(chan FlagsRequestData, 1)
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		switch {
+		case strings.HasPrefix(r.URL.Path, "/flags/definitions") || strings.HasPrefix(r.URL.Path, "/api/feature_flag/local_evaluation"):
+			w.Write([]byte(fixture("feature_flag/test-multiple-flags-valid.json")))
+		case strings.HasPrefix(r.URL.Path, "/flags"):
+			remoteCalls.Add(1)
+			var request FlagsRequestData
+			if err := json.NewDecoder(r.Body).Decode(&request); err != nil {
+				t.Errorf("decode /flags request: %v", err)
+			}
+			remoteRequests <- request
+			w.Write([]byte(`{
+				"flags": {
+					"beta-feature": {"key": "beta-feature", "enabled": false, "metadata": {"id": 1, "version": 1}},
+					"remote-only": {"key": "remote-only", "enabled": true, "metadata": {"id": 2, "version": 1}},
+					"unrequested": {"key": "unrequested", "enabled": true, "metadata": {"id": 3, "version": 1}}
+				}
+			}`))
+		}
+	}))
+	t.Cleanup(server.Close)
+
+	client, _, _ := newEvalClient(t, server, func(c *Config) {
+		c.PersonalApiKey = "personal-key"
+	})
+	waitForFlagDefinitions(t, client)
+
+	flagKeys := []string{"beta-feature", "remote-only"}
+	localSnapshot, err := client.EvaluateFlags(EvaluateFlagsPayload{
+		DistinctId:          "user-1",
+		FlagKeys:            flagKeys,
+		OnlyEvaluateLocally: true,
+	})
+	if err != nil {
+		t.Fatalf("local-only EvaluateFlags error: %v", err)
+	}
+	if got := localSnapshot.Keys(); len(got) != 1 || got[0] != "beta-feature" {
+		t.Fatalf("expected only the locally resolved flag, got %v", got)
+	}
+	if got := remoteCalls.Load(); got != 0 {
+		t.Fatalf("local-only evaluation made %d /flags requests", got)
+	}
+
+	snapshot, err := client.EvaluateFlags(EvaluateFlagsPayload{
+		DistinctId: "user-1",
+		FlagKeys:   flagKeys,
+	})
+	if err != nil {
+		t.Fatalf("EvaluateFlags error: %v", err)
+	}
+	if got := remoteCalls.Load(); got != 1 {
+		t.Fatalf("expected one /flags fallback, got %d", got)
+	}
+	request := <-remoteRequests
+	if len(request.FlagKeysToEvaluate) != 2 || request.FlagKeysToEvaluate[0] != "beta-feature" || request.FlagKeysToEvaluate[1] != "remote-only" {
+		t.Fatalf("expected the original requested keys, got %v", request.FlagKeysToEvaluate)
+	}
+	if got := snapshot.Keys(); len(got) != 2 || got[0] != "beta-feature" || got[1] != "remote-only" {
+		t.Fatalf("expected the snapshot to stay scoped to requested keys, got %v", got)
+	}
+	if !snapshot.IsEnabled("beta-feature") {
+		t.Fatal("expected the locally resolved beta-feature value to override the remote value")
+	}
+	if !snapshot.flags["beta-feature"].LocallyEvaluated {
+		t.Fatal("expected beta-feature to remain locally evaluated")
+	}
+	if !snapshot.IsEnabled("remote-only") {
+		t.Fatal("expected remote-only to be filled by /flags")
+	}
+	if snapshot.flags["remote-only"].LocallyEvaluated {
+		t.Fatal("expected remote-only to be marked as remotely evaluated")
+	}
+}
+
 func TestEvaluateFlags_EmptyDistinctId_NoEvents(t *testing.T) {
 	t.Parallel()
 	var calls atomic.Int32
--- a/feature_flag_evaluations_test.go
+++ b/feature_flag_evaluations_test.go
@@ -657,6 +657,45 @@ func TestEvaluateFlags_MissingRequestedLocalFlagFallsBackRemotely(t *testing.T) 
 	}
 }
 
+func TestEvaluateFlags_MissingRequestedFlagAbsentRemotelyRequestsEachCall(t *testing.T) {
+	t.Parallel()
+	var remoteCalls atomic.Int32
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		switch {
+		case strings.HasPrefix(r.URL.Path, "/flags/definitions") || strings.HasPrefix(r.URL.Path, "/api/feature_flag/local_evaluation"):
+			w.Write([]byte(fixture("feature_flag/test-multiple-flags-valid.json")))
+		case strings.HasPrefix(r.URL.Path, "/flags"):
+			remoteCalls.Add(1)
+			w.Write([]byte(`{"flags": {}}`))
+		}
+	}))
+	t.Cleanup(server.Close)
+
+	client, _, _ := newEvalClient(t, server, func(c *Config) {
+		c.PersonalApiKey = "personal-key"
+		c.DefaultFeatureFlagsPollingInterval = time.Hour
+	})
+	waitForFlagDefinitions(t, client)
+
+	payload := EvaluateFlagsPayload{
+		DistinctId: "user-1",
+		FlagKeys:   []string{"never-defined"},
+	}
+	for i := 0; i < 2; i++ {
+		snapshot, err := client.EvaluateFlags(payload)
+		if err != nil {
+			t.Fatalf("EvaluateFlags call %d error: %v", i+1, err)
+		}
+		if got := snapshot.Keys(); len(got) != 0 {
+			t.Fatalf("expected missing flag to stay absent, got %v", got)
+		}
+	}
+
+	if got := remoteCalls.Load(); got != 2 {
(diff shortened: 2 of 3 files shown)


dustinbyrne (approved):

posthog.go
```diff
@@ -1334,7 +1350,7 @@ func (c *client) populateLocalEvaluations(records map[string]evaluatedFlagRecord
 		locallyEvaluated[storedFlag.Key] = struct{}{}
 	}
 
-	return fallbackToRemote
+	return fallbackToRemote || len(missingFlagKeys) > 0
```
haacked: `suggestion`: A key missing from local definitions forces a `/flags` round trip on every `EvaluateFlags` call. At 500 req/s, that becomes 500 outbound requests per second that resolve nothing, and nothing in the logs points at which key is doing it.

The other branch that forces this same fallback does log. When a flag can't be computed locally, [posthog.go:1313](https://github.com/PostHog/posthog-go/blob/e052d5d164eff0c85200bca4d9dfe782e1e715c0/posthog.go#L1313) logs the key and the reason. The missing-key path reaches the same outcome silently, leaving an operator who notices unexpected `/flags` traffic with no way to find the stale key.

Matching that existing `debugf` is a few lines. `sort` and `strings` are already imported in this file.

```suggestion
	if len(missingFlagKeys) > 0 {
		missing := make([]string, 0, len(missingFlagKeys))
		for key := range missingFlagKeys {
			missing = append(missing, key)
		}
		sort.Strings(missing)
		c.debugf("Requested flags [%s] are absent from local definitions - falling back to /flags", strings.Join(missing, ", "))
		return true
	}
	return fallbackToRemote
```

Use `debugf` rather than a warning on purpose, since a permanently missing key hits this on every call. If you want to bound the cost and not just diagnose it, remembering which keys `/flags` also failed to return (scoped to the poller's definitions generation so a newly created flag starts resolving after the next reload) would cap it at one request per unknown key per poll cycle.

.changeset/quiet-flags-fallback.md
```diff
@@ -0,0 +1,5 @@
+---
+"posthog-go": minor
+---
+
+Fall back to `/flags` when a requested flag is missing from loaded local definitions. This changes the earlier behavior where the key was omitted without a request.
```
haacked: `suggestion`: The release note covers the fallback but not the concrete consequences a caller would see on upgrade, so nobody reading it will know to audit their `FlagKeys` lists.

It doesn't say the returned snapshot is now scoped to `FlagKeys`, so keys the server returns beyond the requested set no longer appear ([posthog.go:1241](https://github.com/PostHog/posthog-go/blob/e052d5d164eff0c85200bca4d9dfe782e1e715c0/posthog.go#L1241)). It also doesn't say that a key resolving neither locally nor remotely now costs a `/flags` request on every call, or that a failed request surfaces as an error on a path that used to be entirely offline.

`minor` is the right bump. The API shape doesn't change, but the body needs to carry the weight.

```suggestion
Fall back to `/flags` when a requested flag key has no loaded local definition. Previously the key was dropped and no request was made.

Two consequences for callers that pass `FlagKeys`:

- A key that resolves neither locally nor remotely triggers a `/flags` request on every `EvaluateFlags` call, since snapshots are not cached. A failed request is returned as an error from calls that previously resolved entirely offline. Set `OnlyEvaluateLocally` to keep the old strictly-local behavior.
- The returned snapshot is now scoped to `FlagKeys`. Keys the server returns beyond the requested set are no longer included, which also means fewer `$feature/<key>` properties on events built from the snapshot via `Capture.Flags`.
```

posthog.go
```diff
@@ -1139,16 +1141,18 @@ type EvaluateFlagsPayload struct {
...
-	// FlagKeys, when non-empty, trims the network call by asking the server
-	// to evaluate only the named flags (sent as flag_keys_to_evaluate).
-	// This is server-side filtering; use FeatureFlagEvaluations.Only to do
-	// client-side filtering of which flags are attached to events from an
-	// existing snapshot.
+	// FlagKeys, when non-empty, restricts local evaluation, the remote request,
+	// and the returned snapshot to the named flags. A requested key missing from
+	// loaded local definitions triggers a /flags fallback unless
+	// OnlyEvaluateLocally is true. EvaluateFlags does not cache snapshots, so a
+	// key that does not exist remotely can trigger a request on each call.
```
haacked: `suggestion`: This comment undersells the cost in two ways, and both matter to someone deciding whether to use `FlagKeys`.

It singles out a key that doesn't exist remotely, but the request goes out on every call for any key missing from local definitions, including one the server resolves fine. Nothing caches the result between calls.

It also doesn't say the fallback can fail. When `/flags` is unreachable, `EvaluateFlags` returns a non-nil error even though every other requested key resolved locally, on a call that made no network request at all before this change. Callers who write `if err != nil { return defaults }` will start discarding a snapshot that still holds live local values.

```suggestion
	// key missing from local definitions triggers a /flags request on every call.
	// A failed request comes back as an error even when every other requested key
	// resolved locally, so read the returned snapshot rather than discarding it.
```

posthog.go
```diff
@@ -1269,8 +1282,10 @@ func (c *client) populateLocalEvaluations(records map[string]evaluatedFlagRecord
 	}
 
 	flagKeyFilter := map[string]struct{}{}
+	missingFlagKeys := map[string]struct{}{}
 	for _, k := range payload.FlagKeys {
 		flagKeyFilter[k] = struct{}{}
+		missingFlagKeys[k] = struct{}{}
```
haacked: `suggestion`: `missingFlagKeys` tracks something the function already knows, and keeping the two in sync costs a mutation buried in the middle of a 55-line loop. Every key that resolves locally is recorded in `locallyEvaluated` at [posthog.go:1350](https://github.com/PostHog/posthog-go/blob/e052d5d164eff0c85200bca4d9dfe782e1e715c0/posthog.go#L1350), so "requested but not resolved locally" is a set difference you can take once after the loop.

The two forms return the same value. A requested key whose definition resolves lands in `locallyEvaluated`. A requested key with no definition lands in neither map. A requested key whose definition fails to compute is deleted from `missingFlagKeys` today but is also absent from `locallyEvaluated`, and that branch already sets `fallbackToRemote` at [posthog.go:1314](https://github.com/PostHog/posthog-go/blob/e052d5d164eff0c85200bca4d9dfe782e1e715c0/posthog.go#L1314), so both forms return true for it.

```diff
 	flagKeyFilter := map[string]struct{}{}
-	missingFlagKeys := map[string]struct{}{}
 	for _, k := range payload.FlagKeys {
 		flagKeyFilter[k] = struct{}{}
-		missingFlagKeys[k] = struct{}{}
 	}
@@
 		if len(flagKeyFilter) > 0 {
 			if _, ok := flagKeyFilter[storedFlag.Key]; !ok {
 				continue
 			}
-			delete(missingFlagKeys, storedFlag.Key)
 		}
@@
-	return fallbackToRemote || len(missingFlagKeys) > 0
+	if fallbackToRemote {
+		return true
+	}
+	// A requested key with no local definition stays unresolved, so ask /flags.
+	for _, key := range payload.FlagKeys {
+		if _, resolved := locallyEvaluated[key]; !resolved {
+			return true
+		}
+	}
+	return false
```

Both new tests should pass unchanged, and the same loop gives you the key list if you also want to log which keys forced the fallback.

feature_flag_evaluations_test.go
```diff
@@ -579,6 +579,123 @@ func TestEvaluateFlags_ForwardsFlagKeys(t *testing.T) {
 	}
 }
 
+func TestEvaluateFlags_MissingRequestedLocalFlagFallsBackRemotely(t *testing.T) {
```
haacked: `suggestion`: The new fallback can set the snapshot to quota-limited on a call that previously never went remote, and nothing tests that path.

A missing `FlagKeys` entry forces a `/flags` request. If that response is quota-limited, `quotaLimited` is set on the whole snapshot at [posthog.go:1234](https://github.com/PostHog/posthog-go/blob/e052d5d164eff0c85200bca4d9dfe782e1e715c0/posthog.go#L1234), and `recordAccess` then stamps `quota_limited` into `$feature_flag_error` for every key accessed on it, including flags that resolved cleanly from local definitions ([feature_flag_evaluations.go:290](https://github.com/PostHog/posthog-go/blob/e052d5d164eff0c85200bca4d9dfe782e1e715c0/feature_flag_evaluations.go#L290)). The stamping itself isn't new, but reaching it from an all-local batch that contains one unknown key is, and no test in this file sets `quota_limited` at all.

Mirror this test's structure with the `/flags` handler returning `{"quota_limited": ["feature_flags"], "flags": {}}`, then assert what a locally-resolved flag like `beta-feature` reports.

feature_flag_evaluations_test.go
```diff
@@ -579,6 +579,123 @@ func TestEvaluateFlags_ForwardsFlagKeys(t *testing.T) {
 	}
 }
```
haacked: `suggestion`: The response filter this PR adds has no coverage for clients without local evaluation, which is the default cloud configuration.

`TestEvaluateFlags_MissingRequestedLocalFlagFallsBackRemotely` only exercises the filter with a poller configured and `locallyEvaluated` non-empty. With no `PersonalApiKey`, `locallyEvaluated` is empty and the `requestedFlagKeys` check at [posthog.go:1241](https://github.com/PostHog/posthog-go/blob/e052d5d164eff0c85200bca4d9dfe782e1e715c0/posthog.go#L1241) is the only thing trimming the response, which is a different state and nothing asserts it.

`TestEvaluateFlags_ForwardsFlagKeys` already sets that scenario up. The [`fixtures/test-flags-v4.json`](https://github.com/PostHog/posthog-go/blob/e052d5d164eff0c85200bca4d9dfe782e1e715c0/fixtures/test-flags-v4.json) file holds 10 flags while the test requests 2, so the difference is directly observable after the change.

Capture the snapshot instead of discarding it, and add:

```go
	if got := snap.Keys(); len(got) != 2 || got[0] != "disabled-flag" || got[1] != "enabled-flag" {
		t.Errorf("expected the snapshot scoped to the requested keys, got %v", got)
	}
```

posthog.go
```diff
@@ -1229,7 +1233,16 @@ func (c *client) evaluateFlagsWithContext(ctx context.Context, payload EvaluateF
 			errorsWhileComputing = flagsResponse.ErrorsWhileComputingFlags
 			quotaLimited = c.isFeatureFlagsQuotaLimited(flagsResponse)
 			if !quotaLimited {
+				requestedFlagKeys := make(map[string]struct{}, len(payload.FlagKeys))
+				for _, key := range payload.FlagKeys {
+					requestedFlagKeys[key] = struct{}{}
+				}
 				for key, detail := range flagsResponse.Flags {
+					if len(requestedFlagKeys) > 0 {
```
haacked: `question`: When only one requested key is missing locally, the fallback still asks `/flags` to evaluate every requested key. Then most of the results get discarded a few lines below at [posthog.go:1246](https://github.com/PostHog/posthog-go/blob/e052d5d164eff0c85200bca4d9dfe782e1e715c0/posthog.go#L1246) because they're already in `locallyEvaluated`.

That's server-side work rather than just extra response bytes. The `flag_keys_to_evaluate` parameter restricts which flags the server clones into evaluation stages and matches, so a caller asking for 20 keys where 19 resolve locally pays for 20 flag evaluations, each with its own cohort and person-property matching, to use one result.

Narrowing the outgoing list to only unresolved keys couldn't send an empty list. Whenever the fallback fires with a non-empty `FlagKeys`, at least one requested key either has no local definition or failed to compute, and neither of those lands in `locallyEvaluated`.

Does sdk-specs#44 mandate forwarding the caller's original list? The PR description calls it deliberate, and both `TestEvaluateFlags_ForwardsFlagKeys` and the new test assert it, so if the spec requires it then this is settled. I mainly want to check that the extra server-side evaluation is a known cost rather than a side effect.

feature_flag_evaluations_test.go
```diff
@@ -579,6 +579,123 @@ func TestEvaluateFlags_ForwardsFlagKeys(t *testing.T) {
...
 
+func TestEvaluateFlags_MissingRequestedLocalFlagFallsBackRemotely(t *testing.T) {
+	t.Parallel()
+	var remoteCalls atomic.Int32
+	remoteRequests := make(chan FlagsRequestData, 1)
```
haacked: `nit`: If a second `/flags` request ever reaches this handler the test hangs instead of failing, because `remoteRequests` holds one value and nothing drains it until after the count assertion, so the second handler goroutine blocks on the send and `server.Close()` waits on it. A package timeout panic then replaces the "expected one /flags fallback" message this test exists to produce.

```suggestion
	remoteRequests := make(chan FlagsRequestData, 4)
```

feature_flag_evaluations_test.go
```diff
@@ -579,6 +579,123 @@ func TestEvaluateFlags_ForwardsFlagKeys(t *testing.T) {
...
+	client, _, _ := newEvalClient(t, server, func(c *Config) {
+		c.PersonalApiKey = "personal-key"
+	})
+	waitForFlagDefinitions(t, client)
+
+	flagKeys := []string{"beta-feature", "remote-only"}
+	localSnapshot, err := client.EvaluateFlags(EvaluateFlagsPayload{
+		DistinctId:          "user-1",
+		FlagKeys:            flagKeys,
+		OnlyEvaluateLocally: true,
```
haacked: `nit`: The only guard on the changeset's "keep `OnlyEvaluateLocally` strictly local" claim lives inside a test named `...FallsBackRemotely`, so it's easy to lose in a future edit to the fallback assertions. Worth extracting this block into its own `TestEvaluateFlags_OnlyEvaluateLocallySkipsMissingKeys`.

PR #299, opened 2026-08-26
fix: match exact/is_not case-insensitively and after string coercion

eeshsaxena:
## Problem

In local flag evaluation, the `exact` and `is_not` operators compare with a plain Go `==`, so they are case-sensitive and type-strict:

```go
if operator == "exact" {
    switch t := value.(type) {
    case []interface{}:
        return contains(t, override_value), nil   // contains also uses ==
    default:
        return value == override_value, nil
    }
}
```

Every other string operator in this file (`icontains`, `starts_with`, `ends_with`) already lower-cases and string-coerces both sides, and the reference SDKs do the same for `exact` (e.g. posthog-python's `str_iequals` compares `str(a).casefold() == str(b).casefold()`). So the same flag can evaluate differently depending on which SDK an app uses:

- `exact "US"` did not match a `"us"` property value
- `exact 1` did not match a `"1"` property value

## Fix

Add an `exactMatch` helper that folds case and coerces to string (matching the other operators and the reference), and route `exact`/`is_not` through it. This replaces the now-unused `contains`.

```go
func exactMatch(value interface{}, override_value interface{}) bool {
    overrideStr := valueToString(override_value)
    if list, ok := value.([]interface{}); ok {
        for _, item := range list {
            if strings.EqualFold(valueToString(item), overrideStr) {
                return true
            }
        }
        return false
    }
    return strings.EqualFold(valueToString(value), overrideStr)
}
```

Same-case, same-type inputs are unchanged, so existing behavior for those is preserved. Added `TestMatchPropertyExactCaseInsensitiveAndCoerced` covering the case-fold and coercion cases (scalar, list, and `is_not`), plus a changeset.

`go build ./...` passes. I wasn't able to finish `go test ./...` locally (slow module downloads in my environment), but I verified the existing `exact` tests all use same-case/same-type values, so they are unaffected.

### DIFF
--- a/.changeset/fix-exact-case-insensitive.md
+++ b/.changeset/fix-exact-case-insensitive.md
@@ -1,0 +1,5 @@
+---
+"posthog-go": patch
+---
+
+Fix `exact`/`is_not` local flag evaluation to match case-insensitively and after string coercion, consistent with the other string operators here and the reference SDKs. A plain Go `==` made them case- and type-sensitive, so e.g. `exact "US"` did not match a `"us"` property value and `exact 1` did not match a `"1"` property value.
--- a/feature_flags_matching_test.go
+++ b/feature_flags_matching_test.go
@@ -161,6 +161,32 @@ func TestMatchPropertySliceExact(t *testing.T) {
 	require.False(t, isMatch)
 }
 
+func TestMatchPropertyExactCaseInsensitiveAndCoerced(t *testing.T) {
+	// "exact"/"is_not" match case-insensitively and after string coercion, like
+	// the icontains/starts_with operators here and the reference SDKs. Plain Go
+	// "==" made them case- and type-sensitive, so these used to be wrong.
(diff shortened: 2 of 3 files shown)


eeshsaxena:
The practical impact is a cross-SDK inconsistency: a flag with an exact condition like country = "US" targets a user whose property is "us" in posthog-python but not in posthog-go today, so the same person can land in different flag variants depending on the SDK. This change brings exact/is_not in line with the other string operators here and with the reference SDKs. Happy to adjust the approach if you would rather fold case a different way.

dustinbyrne:
thanks @eeshsaxena! this looks great. do you mind enabling [commit signing](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) and rebase / force push to sign your commit? it's required to merge to main.

once that's done it looks like we can merge this assuming CI passes 🚀

marandaneto (approved):

PR #300, opened 2026-08-27
fix(flags): align local exact matching with the flags service

marandaneto:

Local feature flag evaluation did not consistently match the released Rust flags service for property filters.

This is part of [PostHog/posthog#78019](https://github.com/PostHog/posthog/issues/78019).

This change:

- folds only ASCII `A-Z` for contains, prefix, and suffix operators
- uses full Unicode lowercasing for `exact` and `is_not`
- applies the released backend's boolean-array precedence, including its current empty-array behavior
- stringifies null, arrays, objects, and representable numbers using the backend's canonical JSON forms
- returns an inconclusive result for integral `float64` filter values whose original JSON integer/float kind was lost
- preserves ordinary ANY membership for non-boolean-like filter arrays
- adds a patch changeset for `posthog-go`

The proposed backend behavior change in [PostHog/posthog#90694](https://github.com/PostHog/posthog/pull/90694) is a separate draft. This PR continues to match the currently released backend until that proposal is accepted and shipped.

### Runtime limitation

Go's normal JSON decoding collapses `323` and `323.0` into the same `float64` filter value. Local evaluation now falls back to the server for ambiguous integral filter numbers instead of guessing. Known string filter spellings and caller-provided integer or floating-point property types still use their canonical forms.

The new focused tests reproduced these failures before the fix:

- boolean-like arrays used ordinary ANY membership instead of aggregate backend truthiness
- empty arrays did not use the backend's vacuous `all()` behavior
- null, arrays, and objects used Go-specific text rather than canonical JSON
- negative zero and exponent forms differed from `serde_json`
- integral decoded filter numbers produced a local answer despite lost type information

After the fix:

- `go test ./...`
- `go test -race . -run 'TestMatchPropertyExact' -count=1`
- `go vet ./...`
- `gofmt` and `git diff --check`
- local autoreview reported no actionable findings after two accepted review fixes for typed containers and recursively ambiguous numbers

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Pi implemented this user-directed compatibility fix using the address-pr-comments, check-pr, autoreview, and karpathy-guidelines skills. The final implementation keeps `x/text/cases.Lower` because verification showed it already matches the backend's final-sigma and dotted-I behavior.

### DIFF
--- a/.changeset/align-flag-case-folding.md
+++ b/.changeset/align-flag-case-folding.md
@@ -2,4 +2,4 @@
 "posthog-go": patch
 ---
 
-Align local feature flag property matching with the flags service: use ASCII-only case folding for contains, prefix, and suffix operators, and Unicode lowercasing rather than broader case folding for `exact` and `is_not`.
+Align local feature flag property matching with the flags service, including boolean-array precedence, canonical JSON stringification, and operator-specific case folding.
--- a/.changeset/align-flag-case-folding.md
+++ b/.changeset/align-flag-case-folding.md
@@ -1,0 +1,5 @@
+---
+"posthog-go": patch
+---
+
+Align local feature flag property matching with the flags service: use ASCII-only case folding for contains, prefix, and suffix operators, and Unicode lowercasing rather than broader case folding for `exact` and `is_not`.
--- a/feature_flags_matching_test.go
+++ b/feature_flags_matching_test.go
@@ -173,9 +173,12 @@ func TestMatchPropertyExactCaseInsensitiveAndCoerced(t *testing.T) {
 	require.True(t, exact("US", "us"))
 	require.True(t, exact("Value", "value"))
 	require.True(t, exact("Ä", "ä"))
+	require.True(t, exact("İ", "i\u0307"))
+	require.True(t, exact("ΟΣ", "ος"))
 	require.True(t, exact(1, "1"))
 	require.True(t, exact("1", 1))
 	require.False(t, exact("US", "CA"))
+	require.False(t, exact("İ", "i"))
 	require.False(t, exact("Σ", "ς"))
 	require.False(t, exact("ß", "ss"))
 
--- a/feature_flags_matching_test.go
+++ b/feature_flags_matching_test.go
@@ -172,19 +172,55 @@ func TestMatchPropertyExactCaseInsensitiveAndCoerced(t *testing.T) {
 	}
 	require.True(t, exact("US", "us"))
 	require.True(t, exact("Value", "value"))
+	require.True(t, exact("Ä", "ä"))
 	require.True(t, exact(1, "1"))
 	require.True(t, exact("1", 1))
 	require.False(t, exact("US", "CA"))
+	require.False(t, exact("Σ", "ς"))
+	require.False(t, exact("ß", "ss"))
 
-	require.True(t, exact([]interface{}{"US", "CA"}, "us"))
-	require.False(t, exact([]interface{}{"US", "CA"}, "gb"))
+	require.True(t, exact([]interface{}{"US", "Ä"}, "ä"))
+	require.False(t, exact([]interface{}{"US", "Σ"}, "ς"))
 
 	isNot, err := matchProperty(FlagProperty{Key: "k", Value: "US", Operator: "is_not"}, NewProperties().Set("k", "us"))
 	require.NoError(t, err)
 	require.False(t, isNot)
-	isNot, err = matchProperty(FlagProperty{Key: "k", Value: "US", Operator: "is_not"}, NewProperties().Set("k", "gb"))
+	isNot, err = matchProperty(FlagProperty{Key: "k", Value: "Σ", Operator: "is_not"}, NewProperties().Set("k", "ς"))
 	require.NoError(t, err)
 	require.True(t, isNot)
+}
+
+func TestMatchPropertyStringOperatorsUseASCIICaseFolding(t *testing.T) {
+	tests := []struct {
+		operator string
+		filter   string
+		actual   string
+		expected bool
+	}{
+		{operator: "icontains", filter: "VALUE", actual: "a-value-b", expected: true},
+		{operator: "not_icontains", filter: "VALUE", actual: "a-value-b", expected: false},
+		{operator: "starts_with", filter: "VALUE", actual: "value-b", expected: true},
+		{operator: "not_starts_with", filter: "VALUE", actual: "value-b", expected: false},
+		{operator: "ends_with", filter: "VALUE", actual: "a-value", expected: true},
+		{operator: "not_ends_with", filter: "VALUE", actual: "a-value", expected: false},
+		{operator: "icontains", filter: "ä", actual: "Ä", expected: false},
+		{operator: "not_icontains", filter: "ä", actual: "Ä", expected: true},
+		{operator: "starts_with", filter: "ä", actual: "Äbc", expected: false},
+		{operator: "not_starts_with", filter: "ä", actual: "Äbc", expected: true},
+		{operator: "ends_with", filter: "ä", actual: "bcÄ", expected: false},
+		{operator: "not_ends_with", filter: "ä", actual: "bcÄ", expected: true},
+	}
+
+	for _, test := range tests {
+		t.Run(test.operator+"/"+test.filter, func(t *testing.T) {
+			matched, err := matchProperty(
+				FlagProperty{Key: "k", Value: test.filter, Operator: test.operator},
+				NewProperties().Set("k", test.actual),
+			)
+			require.NoError(t, err)
+			require.Equal(t, test.expected, matched)
+		})
+	}
 }
 
 func TestMatchPropertyNumber(t *testing.T) {
--- a/feature_flags_matching_test.go
+++ b/feature_flags_matching_test.go
@@ -2,12 +2,21 @@ package posthog
 
 import (
 	"errors"
+	"math"
 	"strconv"
 	"testing"
 	"time"
 
 	"github.com/stretchr/testify/require"
 )
+
+type exactNamedString string
+
+type exactMarshaledString struct{}
+
+func (exactMarshaledString) MarshalJSON() ([]byte, error) {
+	return []byte(`"pro"`), nil
+}
 
 func TestMatchPropertyValue(t *testing.T) {
 	property := FlagProperty{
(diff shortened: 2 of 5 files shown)


ioannisj (approved):

arnohillen:
`valueToString` (`featureflags.go:1841`) falls through to `fmt.Sprint` for composite values, so an array property `[1,2]` stringifies to `[1 2]` and an object to `map[a:1]` where the service's `to_string_representation` (`property_matching.rs:54-62`) gives `[1,2]` and `{"a":1}` (executed both sides: `exact` filter `"[1,2]"` against that array property returns `false` here, `true` on the service); pre-existing on `main` rather than introduced here, but it contradicts the new comment's claim that `exactMatch` stringifies the way the flags service does.

arnohillen:
`exactMatch` (`featureflags.go:1760`) has no equivalent of the service's boolean-coercion branch (`property_matching.rs:212-219`), so `exact` filter `["true","false"]` against person property `true` matches here and does not on the service, and `exact` filter `["false"]` against `"banana"` does not match here and does on the service; pre-existing (identical at `749f1e8`) rather than introduced by this PR, and shared with posthog-python.

marandaneto:
Addressed both review concerns in signed commit `a14b017`.

- Comment `5443082062` (composite stringification): reproduced `"[1,2]"` versus `[1,2]` as a false local result. The same focused test now passes for null, compact arrays, recursively sorted objects, negative zero, `1e-7`, `1e+16`, `0.00001`, and `0.000099`. Typed numeric containers preserve integer spellings.
- Comment `5443083104` (boolean precedence): reproduced `["true","false"]` versus `true` as a local match where the released backend does not match, plus the inverse `"pro"` case and empty-array cases. The table-driven exact/is_not matrix now passes with the backend boolean gate before ordinary array membership.

Integral `float64` filter values remain lexically ambiguous after JSON decoding, so local evaluation now returns `InconclusiveMatchError` and uses the existing server fallback rather than guessing between `323` and `323.0`. `go test ./...`, focused race tests, `go vet ./...`, formatting, and autoreview pass. PostHog/posthog#90694 remains a separate draft proposal; this PR matches the released backend.

featureflags.go
```diff
@@ -1827,6 +1970,145 @@ func valueToString(v interface{}) string {
...
+	}
+}
+
+func serdeFloatString(value float64, bitSize int) (string, bool) {
+	if math.IsNaN(value) || math.IsInf(value, 0) {
+		return "", false
+	}
+
+	absolute := math.Abs(value)
+	if absolute != 0 && (absolute < 1e-6 || absolute >= 1e16) {
```
haacked: This threshold is one decade off, and there is a second float rule below it that also disagrees with the service.

serde_json switches to scientific notation below `1e-5`, not below `1e-6`. I checked the source rather than inferring it: `rust/Cargo.lock` pins serde_json 1.0.149, whose `write_f64` goes through `zmij::Buffer::format_finite`, and zmij uses fixed notation for f64 only when the decimal exponent is in `-5..=15`. ryu, which older serde_json used, agrees.

So a filter of `5e-6` renders as `0.000005` here and as `5e-6` on the service. The strings compare unequal, the match returns false, and no error is raised, so nothing falls back.

```suggestion
	if absolute != 0 && (absolute < 1e-5 || absolute >= 1e16) {
```

The second rule is the `.0` suffix. It is correct when the value's JSON spelling carried a `.` or an `e`, and wrong when Go writes the value as an integer. `json.Marshal(float64(20))` is `20`, which serde_json reads as an integer and prints as `20`, so the service says `20` where this says `20.0`. Same for `float64(1e16)`, which Go sends as `10000000000000000`.

I ran serde_json on the text Go emits to get the rule exactly. It reads a JSON number as an integer only when the text has no `.`, `e`, or `E`, is not `-0`, and fits i64/u64:

```go
// wireFloatString renders a Go float the way the flags service will after the SDK
// marshals it. serde_json reads a JSON number as an integer only when the text has
// no fraction or exponent, is not "-0", and fits i64/u64.
func wireFloatString(value float64, bitSize int) (string, bool) {
	encoded, err := json.Marshal(value)
	if err != nil {
		return "", false
	}
	text := string(encoded)
	if !strings.ContainsAny(text, ".eE") && text != "-0" && fitsJSONInteger(text) {
		return text, true
	}
	return serdeFloatString(value, bitSize)
}

func fitsJSONInteger(text string) bool {
	if _, err := strconv.ParseInt(text, 10, 64); err == nil {
		return true
	}
	_, err := strconv.ParseUint(text, 10, 64)
	return err == nil
}
```

with `canonicalJSONValue`'s `float32` and `float64` cases calling it instead. The `-0` guard is load-bearing: Go writes negative zero as `-0`, serde_json parses that as a float and prints `-0.0`, so your existing `negative zero` row is already right and should stay.

With both changes I compared 22 values against serde_json and got no mismatches. Three test rows move: `integral float` becomes `filter: "323"`, `large exponent` becomes `filter: "10000000000000000"`, and the last assertion in `TestMatchPropertyExactFallsBackForAmbiguousFilterNumbers` becomes `require.True`. Worth adding a row inside the broken decade too, since the current table brackets it at `1e-7` and `0.00001` without landing in it:

```go
		{name: "lower scientific boundary", filter: "1e-6", actual: 1e-6},
```

featureflags.go
```diff
@@ -1750,24 +1759,158 @@ func interfaceToFloat(val interface{}) (float64, error) {
...
+				return true, nil
+			}
+		}
+		if hadAmbiguousNumber {
+			return false, errAmbiguousExactNumber
+		}
+		return false, nil
+	}
+
+	if containsAmbiguousExactNumber(value) {
```
haacked: This makes every whole-number `exact` filter stop evaluating locally, which is a bigger blast radius than the "runtime limitation" note suggests.

`containsAmbiguousExactNumber` returns true for any integral `float64`, and the definitions response is decoded at `featureflags.go:580` with no `UseNumber()`, so a filter written as `25` always arrives as `float64(25)`. The check runs before any comparison, so it fires even when neither spelling could have matched.

I ran it end to end on both revisions, serving `{"key":"age","operator":"exact","value":25}` and evaluating against `age: 25`. On `749f1e8` it returns `true` with no remote call. Here, `OnlyEvaluateLocally: true` returns nil plus "Can't determine if feature flag is enabled or not with given properties", and without that option it makes a `/flags` call on every evaluation. For array filters it is conditional rather than absolute: an unambiguous element that matches still wins.

I checked how often this can actually happen, since the fixtures are all strings. The server permits and preserves numeric values here. `filters_schema.py` declares `value` as an unconstrained `JSONField` while neighbouring fields are strict-typed, `check_operator_value_compatibility` requires strings for `regex`, `icontains`, `gt`, and friends but not for `exact` or `is_not`, and migration `0011_clean_flag_filters_recoverable_violations` stringifies numeric values while deliberately excluding `exact` and `is_not`. So production rows were swept and numeric `exact` values were left as numbers on purpose. The UI does not emit them, so the realistic sources are the API, terraform, and older flags.

Worth noting that posthog-python, posthog-dotnet, and the Rust service all answer this case locally rather than deferring, so this would make posthog-go the only one that declines.

The information being reconstructed is thrown away by a decode this library owns. Decoding `FlagProperty.Value` with `UseNumber()` keeps `json.Number("25")` and `json.Number("25.0")` distinct, which `canonicalJSONValue` already handles correctly. That makes `containsAmbiguousExactNumber`, `errAmbiguousExactNumber`, and the `hadAmbiguousNumber` bookkeeping deletable.

Three things need to come along with it, so it is a four-line change rather than a one-line one:

- `interfaceToFloat` needs a `json.Number` case, otherwise `gt`/`lt`/`gte`/`lte` break. I confirmed a `gt` filter of `18` against `age: 21` matches today and returns "value 1 is not orderable" after the switch.
- `exactValueToString`'s fast path returns early for anything whose reflect kind is String, and `json.Number` has that kind, so it needs an exclusion or `"2.50"` returns `2.50` where canonical form gives `2.5`.
- `FlagProperty.Value` is exported, so the runtime type flip is visible to anyone type-asserting `float64` on it.

That last one is a real cost and it is your call. If you would rather keep the current approach, the narrower version is to compare both candidate spellings and only fall back when they disagree, which keeps a local answer for the common case where a filter of `323` cannot match `"pro"` under either spelling.

featureflags.go
```diff
@@ -1750,24 +1759,158 @@ func interfaceToFloat(val interface{}) (float64, error) {
...
+			field := reflected.Field(index)
+			if field.CanInterface() && containsAmbiguousExactNumber(field.Interface()) {
+				return true
+			}
+		}
+	}
+	return false
+}
+
+func unicodeLower(value string) string {
```
haacked: Building a caser per call puts about 265 bytes of garbage on a path that previously allocated nothing.

Measured on the pinned x/text v0.21.0: 146.7 ns and 265 B across 1 allocation per call, against 2.5 ns and zero allocations for the `strings.EqualFold` version on `749f1e8`. `exactMatch` calls this two to three times per evaluation, and `isTruthyOrFalsyPropertyValue` calls it again for every string element of an array filter. A single flag check is only a couple of microseconds, so what shows up is allocation rate rather than latency, and it scales with flags times conditions times request rate inside the caller's server.

An ASCII fast path handles it. Every character full Unicode lowercasing treats specially is non-ASCII, so a pure-ASCII string folds identically either way. That was checked over every 1-, 2-, and 3-byte ASCII string and a million random ones on v0.21.0, v0.27.0, and v0.33.0, with no mismatches. Your `falſe` test is unaffected, since U+017F is two bytes both above 0x7F and takes the caser branch.

```go
func unicodeLower(value string) string {
	for i := 0; i < len(value); i++ {
		c := value[i]
		if c >= 0x80 {
			// Casers may be stateful and cannot be shared between evaluator goroutines.
			return cases.Lower(language.Und).String(value)
		}
		if c >= 'A' && c <= 'Z' {
			return asciiLower(value)
		}
	}
	return value
}
```

Returning `value` when nothing needs changing also skips the copy `asciiLower` would make, which is the common case for property values.

Two nearby alternatives do not work, so worth ruling them out explicitly. `strings.ToLower` breaks the parity this PR exists for. A package-level caser is a data race for the reason your comment already gives, and it only recovers about 30% anyway, since the transform dominates rather than the construction.

featureflags.go
```diff
@@ -1158,40 +1164,43 @@ func matchProperty(property FlagProperty, properties Properties) (bool, error) {
...
+		return matched, nil
 	}
 
 	if operator == "is_set" {
 		return true, nil
 	}
 
 	if operator == "icontains" {
-		return strings.Contains(strings.ToLower(valueToString(override_value)), strings.ToLower(valueToString(value))), nil
+		return strings.Contains(asciiLower(valueToString(override_value)), asciiLower(valueToString(value))), nil
```
haacked: These six operators got the ASCII folding fix but still call `valueToString`, so the same property now stringifies two different ways depending on which operator reads it.

`valueToString` renders an array property as `[1 2]` via `fmt.Sprint` and collapses `323.0` to `323`. That is the same bug @arnohillen reported against `exact`, which this PR fixed only on the `exact` side. So an `icontains` filter of `"[1,2]"` against an array property still fails locally and passes on the service.

The reference uses one helper for both groups: `to_string_representation` at `property_matching.rs:54-62` is called from the Exact arm and the Icontains arm alike.

```suggestion
		return strings.Contains(asciiLower(exactValueToString(override_value)), asciiLower(exactValueToString(value))), nil
```

and the same on the other five. Worth leaving `valueToString` in place for the cohort ID lookup, which is keying rather than matching. If you do unify them, `exactValueToString` stops being exact-specific and reads better as `toStringRepresentation`.

The PR title scopes this to exact matching, so is leaving the other operators on the old helper deliberate follow-up work, or just not reached yet?

go.mod
```diff
@@ -14,6 +14,7 @@ require (
 	github.com/stretchr/testify v1.11.1
 	github.com/urfave/cli v1.22.17
 	golang.org/x/sys v0.21.0
+	golang.org/x/text v0.21.0
```
haacked: This version is inside the advisory range for GO-2026-5970, and govulncheck reports it as called rather than informational, tracing to the line this PR adds: `featureflags.go:1903: posthog.unicodeLower calls cases.Caser.String, which eventually calls norm.Form.Properties`. Any consumer who evaluates flags locally and gates CI on govulncheck gets a red build attributed to this SDK.

It does not look exploitable. The infinite loop is in `nextComposed` in `unicode/norm/iter.go`, and the `cases` package never uses the norm iterator; a thousand malformed-UTF-8 inputs through `cases.Lower` all returned promptly. So this is scanner and supply-chain noise, not a reachable hang.

The upgrade is not free either. The fix landed in v0.39.0, and every release from there on declares `go 1.25.0`, while this module declares `go 1.21` and CI builds on 1.21. Taking it means dropping Go 1.21 support for everyone importing posthog-go.

Both options seem defensible to me, so flagging it as a decision rather than a fix: take the bump and drop 1.21, or stay here and accept that consumers' scanners will point at this line.

.changeset/align-flag-case-folding.md
```diff
@@ -0,0 +1,5 @@
+---
+"posthog-go": patch
```
haacked: Two shipped behaviors change here, and a `patch` sends them through `go get -u` and Dependabot automerge without describing either.

`icontains`, `starts_with`, and `ends_with` lose Unicode case folding. An `icontains` filter of `MÜNCHEN` against a `münchen` property returns true on v1.24.2 and false here. And `exact` with a filter like `["true", "false"]` now compares aggregate truthiness rather than testing membership, so it returns false against `"true"` where it used to return true.

Both are intended parity fixes, so this is about the release level and the wording, not the behavior. This repo has used `minor` for local-evaluation behavior changes before, in v1.24.0 and v1.23.0.

```suggestion
"posthog-go": minor
```

Worth expanding the description to name what moves, since this is what SDK users read:

> Align local feature flag property matching with the flags service. Three behaviors change. `icontains`, `starts_with`, and `ends_with` now fold only ASCII `A-Z`, so a filter like `MÜNCHEN` no longer matches a `münchen` property locally, matching the service. `exact` and `is_not` now apply the service's boolean precedence, so a filter of `["true", "false"]` compares aggregate truthiness rather than testing membership. `exact` and `is_not` stringify null, arrays, objects, and numbers as canonical JSON, so an array property `[1,2]` matches the filter `"[1,2]"` instead of `"[1 2]"`.

The pending `.changeset/fix-exact-case-insensitive.md` from #299 is unreleased and describes behavior this PR then rewrites, so it may be worth folding into this one entry rather than shipping both.

feature_flags_matching_test.go
```diff
@@ -172,21 +181,168 @@ func TestMatchPropertyExactCaseInsensitiveAndCoerced(t *testing.T) {
...
+		{name: "false matches arbitrary falsey string", filter: false, actual: "banana", expected: true},
+		{name: "false string matches zero", filter: "false", actual: 0, expected: true},
+		{name: "boolean parsing uses lowercase not case folding", filter: "falſe", actual: false, expected: false},
+		{name: "false array matches null", filter: []interface{}{"false"}, actual: nil, expected: true},
+		{name: "mixed boolean array is aggregate false", filter: []interface{}{"true", "false"}, actual: "pro", expected: true},
+		{name: "mixed boolean array does not use any", filter: []interface{}{"true", "false"}, actual: "true", expected: false},
+		{name: "empty array is aggregate true for boolean", filter: []interface{}{}, actual: true, expected: true},
+		{name: "empty array is aggregate true for string", filter: []interface{}{}, actual: "TRUE", expected: true},
+		{name: "empty array is aggregate true for empty array", filter: []interface{}{}, actual: []interface{}{}, expected: true},
+		{name: "empty array is aggregate true for true array", filter: []interface{}{}, actual: []interface{}{true}, expected: true},
```
haacked: All four empty-array rows expect true, so nothing pins down that an empty filter array fails to match an ordinary value. `[]interface{}{}` against `"us"` returns false today, but a change that made an empty filter array match everything would pass all four rows, and a condition meant to match nobody would match everyone.

The Rust suite keeps this as `test_match_properties_exact_empty_array_never_matches`. One more row covers it, and the table gives you the `is_not` direction for free:

```suggestion
		{name: "empty array is aggregate true for true array", filter: []interface{}{}, actual: []interface{}{true}, expected: true},
		{name: "empty array does not match an ordinary value", filter: []interface{}{}, actual: "us", expected: false},
```

PR #305, opened 2026-08-28
fix: not_regex should handle all value types like regex

eeshsaxena:
## Problem

The `regex` and `not_regex` operators handle value types differently. `regex` coerces both sides with `valueToString`, so it accepts strings, ints, floats, and bools:

```go
if operator == "regex" {
    r, err := getOrCompileRegex(valueToString(value))
    if err != nil { return false, nil }
    return r.MatchString(valueToString(override_value)), nil
}
```

`not_regex` instead uses a manual `string`/`int` type switch and returns an error for anything else:

```go
if valueString, ok := override_value.(string); ok {
    match = r.MatchString(valueString)
} else if valueInt, ok := override_value.(int); ok {
    match = r.MatchString(strconv.Itoa(valueInt))
} else {
    return false, errors.New("value type not supported")
}
```

The important gap is `float64`: JSON numbers deserialize to `float64`, so a `not_regex` condition on a numeric property value errors out even though the same value works with `regex`:

```
regex     ^1  vs float64(123)  ->  (true,  <nil>)
not_regex ^1  vs float64(123)  ->  (false, "value type not supported")   ❌
```

## Fix

Make `not_regex` mirror `regex`, coercing both the pattern and the property value with `valueToString`. This handles all value types and keeps the two operators consistent. Added `TestMatchPropertyNotRegexHandlesAllValueTypes` (string, int, float64).

## Testing

`go test .` passes (full package suite, including the new test).

### DIFF
--- a/.changeset/fix-not-regex-value-types.md
+++ b/.changeset/fix-not-regex-value-types.md
@@ -2,4 +2,4 @@
 "posthog-go": patch
 ---
 
-Fix `not_regex` local flag evaluation erroring on non-string/int property values. The `not_regex` operator used a manual string/int type switch and returned an error for other types, most notably `float64`, which is what JSON numbers deserialize to, so it failed on a numeric property value even though `regex` handled it. `not_regex` now coerces both sides with `valueToString`, mirroring `regex`.
+Fix `not_regex` local flag evaluation erroring on non-string/int property values. The `not_regex` operator used a manual string/int type switch and returned an error for other types, most notably `float64`, which is what JSON numbers deserialize to, so it failed on a numeric property value even though `regex` handled it. `not_regex` now coerces both sides with `valueToString`, mirroring `regex`. It also no longer coerces a `nil` property value to the Go-specific `<nil>` string; like the server, a null property now yields no-match for `not_regex`.
--- a/.changeset/fix-not-regex-value-types.md
+++ b/.changeset/fix-not-regex-value-types.md
@@ -2,4 +2,4 @@
 "posthog-go": patch
 ---
 
-Fix `not_regex` local flag evaluation erroring on non-string/int property values. The `not_regex` operator used a manual string/int type switch and returned an error for other types, most notably `float64`, which is what JSON numbers deserialize to, so it failed on a numeric property value even though `regex` handled it. `not_regex` now coerces both sides with `valueToString`, mirroring `regex`. It also no longer coerces a `nil` property value to the Go-specific `<nil>` string; like the server, a null property now yields no-match for `not_regex`.
+Fix `not_regex` local flag evaluation erroring on non-string/int property values. The `not_regex` operator used a manual string/int type switch and returned an error for other types, most notably `float64`, which is what JSON numbers deserialize to, so it failed on a numeric property value even though `regex` handled it. `not_regex` now coerces both sides with `valueToString`, mirroring `regex`. An explicit `nil` property value is represented as `null`, matching the feature flags evaluation service rather than Go's default `<nil>` representation.
--- a/.changeset/fix-not-regex-value-types.md
+++ b/.changeset/fix-not-regex-value-types.md
@@ -1,0 +1,5 @@
+---
+"posthog-go": patch
+---
+
+Fix `not_regex` local flag evaluation erroring on non-string/int property values. The `not_regex` operator used a manual string/int type switch and returned an error for other types, most notably `float64`, which is what JSON numbers deserialize to, so it failed on a numeric property value even though `regex` handled it. `not_regex` now coerces both sides with `valueToString`, mirroring `regex`.
--- a/feature_flags_matching_test.go
+++ b/feature_flags_matching_test.go
@@ -309,6 +309,20 @@ func TestMatchPropertyRegex(t *testing.T) {
 	}
 }
 
+func TestMatchPropertyNotRegexHandlesAllValueTypes(t *testing.T) {
+	// not_regex must coerce like regex does. JSON numbers arrive as float64,
+	// which the old manual type switch rejected with an error even though
+	// regex handled it.
+	for _, ov := range []interface{}{"123", 123, 123.0, float64(123)} {
+		m, err := matchProperty(FlagProperty{Key: "k", Value: "^1", Operator: "not_regex"}, NewProperties().Set("k", ov))
+		require.NoError(t, err)
+		require.False(t, m) // "123" matches ^1, so not_regex is false
+	}
+	m, err := matchProperty(FlagProperty{Key: "k", Value: "^9", Operator: "not_regex"}, NewProperties().Set("k", 123.0))
+	require.NoError(t, err)
+	require.True(t, m) // "123" does not match ^9
+}
+
 func TestMatchPropertyContains(t *testing.T) {
 	shouldMatch := []interface{}{"value", "value2", "value3", "value4", "343tfvalue5"}
 
--- a/feature_flags_matching_test.go
+++ b/feature_flags_matching_test.go
@@ -322,17 +322,19 @@ func TestMatchPropertyNotRegexHandlesAllValueTypes(t *testing.T) {
 	require.NoError(t, err)
 	require.True(t, m) // "123" does not match ^9
 
-	// A nil property value must not be coerced to Go's "<nil>" string. Like the
-	// server (posthog-python only allows None for "is_not"), a regex operator
-	// yields no-match for a null property.
+	// The feature flags evaluation service stringifies a null property as "null"
+	// before matching it. not_regex must negate the regex result as usual.
 	m, err = matchProperty(FlagProperty{Key: "k", Value: "^1", Operator: "not_regex"}, NewProperties().Set("k", nil))
 	require.NoError(t, err)
-	require.False(t, m)
-	// Regression guard: a pattern that would match the literal "<nil>" must not
-	// sneak through for a null property.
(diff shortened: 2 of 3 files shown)


eeshsaxena:
This pairs with the exact/is_not consistency fix in #299 and is independent of it. The practical trigger is common: a not_regex flag condition on a numeric property (an ID, a version code, an age) that arrives as a JSON number, which becomes float64 in Go, currently errors instead of evaluating. Routing through valueToString keeps not_regex in step with regex for every value type.

featureflags.go
```diff
@@ -1205,32 +1215,18 @@ func matchProperty(property FlagProperty, properties Properties) (bool, error) {
...
-			return false, errors.New("regex expression not allowed")
-		}
-
-		r, err := getOrCompileRegex(pattern)
+		// Mirror the "regex" branch above: coerce both sides with valueToString
+		// so all property/flag value types are handled. The previous manual
+		// string/int type switch errored on other types, most notably float64,
+		// which is what JSON numbers deserialize to, so not_regex failed on a
+		// numeric property value even though regex handled it.
+		r, err := getOrCompileRegex(valueToString(value))
```
dustinbyrne, line 1229: a bit of an edge case, but since we're in here, one minor deviation from the feature flags evaluation service is if `value` is `nil`, the string representation returned is `<nil>`

e.g.  with the following properties:

 ```json
   { "k": null }
 ```

 and condition:

 ```json
   {
     "key": "k",
     "operator": "not_regex",
     "value": "^null$"
   }
 ```

this would match remotely, but not match here.

dustinbyrne (commented):
hey @eeshsaxena, thanks for another pull request here! i ran an agentic review for compatibility, and it had one minor finding i've described below. it's small enough in scope that it seems worth fixing.

featureflags.go
```diff
@@ -1215,32 +1215,18 @@ func matchProperty(property FlagProperty, properties Properties) (bool, error) {
...
-		if valueString, ok := value.(string); ok {
-			pattern = valueString
-		} else if valueInt, ok := value.(int); ok {
-			pattern = strconv.Itoa(valueInt)
-		} else {
-			return false, errors.New("regex expression not allowed")
-		}
-
-		r, err := getOrCompileRegex(pattern)
+		// Mirror the "regex" branch above: coerce both sides with valueToString
```
haacked: `suggestion`: This comment describes the string/int type switch the PR deletes, so once merged it will explain code that is no longer there. It also duplicates what the two lines below it already show.

Deleting it keeps the style consistent with [the `regex` branch above](https://github.com/PostHog/posthog-go/blob/f8425e91b7430a818fd726fc0fc9e0b30749d566/featureflags.go#L1207-L1214), which carries only `// invalid regex`. The changeset already records why the coercion changed.

featureflags.go
```diff
@@ -1215,32 +1215,18 @@ func matchProperty(property FlagProperty, properties Properties) (bool, error) {
...
-		if valueString, ok := override_value.(string); ok {
-			match = r.MatchString(valueString)
-		} else if valueInt, ok := override_value.(int); ok {
-			match = r.MatchString(strconv.Itoa(valueInt))
-		} else {
-			return false, errors.New("value type not supported")
-		}
-
-		return !match, nil
+		return !r.MatchString(valueToString(override_value)), nil
```
haacked: `suggestion`: Building on @dustinbyrne's `nil` note, [`valueToString`](https://github.com/PostHog/posthog-go/blob/f8425e91b7430a818fd726fc0fc9e0b30749d566/featureflags.go#L1810) falls through to `fmt.Sprint` for `nil`, slices, and maps, so this SDK compares against `<nil>` and `[a b]` where the flags service compares against `null` and `["a","b"]`. Before this PR, the old branch rejected these types with `errors.New("value type not supported")`, and that error reaches [`GetFeatureFlag`](https://github.com/PostHog/posthog-go/blob/f8425e91b7430a818fd726fc0fc9e0b30749d566/featureflags.go#L643), which calls `/flags` when `OnlyEvaluateLocally` is false. Now the SDK answers them locally with a different string format.

The coercion belongs in one helper both branches call, so `regex` and `not_regex` are not re-split by the fix. If you would rather keep this PR to the `float64` fix, a follow-up covering both regex branches is a clean split; worth saying so on the thread so the `nil` behavior is a decision rather than a side effect.

feature_flags_matching_test.go
```diff
@@ -309,6 +309,20 @@ func TestMatchPropertyRegex(t *testing.T) {
 	}
 }
 
+func TestMatchPropertyNotRegexHandlesAllValueTypes(t *testing.T) {
```
haacked: `suggestion`: The test varies the property value but never the pattern, so it misses the other half of the change. `value` also moved from the type switch to [`valueToString`](https://github.com/PostHog/posthog-go/blob/f8425e91b7430a818fd726fc0fc9e0b30749d566/featureflags.go#L1223), and both cases here pass a string pattern (`"^1"` and `"^9"`) that the old code handled identically. You could revert the pattern side to the type switch and this test would still pass.

[`FlagProperty.Value`](https://github.com/PostHog/posthog-go/blob/f8425e91b7430a818fd726fc0fc9e0b30749d566/featureflags.go#L177) is `interface{}` with a `json:"value"` tag, so a condition carrying `"value": 4` deserializes to `float64(4)`, which the old branch rejected with "regex expression not allowed". An `int` pattern would not catch it, since the old switch accepted `int`. The ["non string value" block in `TestMatchPropertyRegex`](https://github.com/PostHog/posthog-go/blob/f8425e91b7430a818fd726fc0fc9e0b30749d566/feature_flags_matching_test.go#L290-L308) is the one to mirror, with `Value: float64(4)`.

The name promises all value types, so it is worth either covering the pattern side or narrowing the name.

feature_flags_matching_test.go
```diff
@@ -309,6 +309,20 @@ func TestMatchPropertyRegex(t *testing.T) {
...
+	// not_regex must coerce like regex does. JSON numbers arrive as float64,
+	// which the old manual type switch rejected with an error even though
+	// regex handled it.
+	for _, ov := range []interface{}{"123", 123, 123.0, float64(123)} {
```
haacked: `nit`: `123.0` and `float64(123)` are the same element here, since an untyped float constant in an `[]interface{}` becomes `float64`, so the loop covers three types rather than four and runs `float64` twice. Swapping one for `int64(123)` picks up the remaining scalar arm of [`valueToString`](https://github.com/PostHog/posthog-go/blob/f8425e91b7430a818fd726fc0fc9e0b30749d566/featureflags.go#L1795-L1813), which formats `int64` with `strconv.FormatInt`, so it still stringifies to "123" and the `require.False` assertion holds.

```suggestion
	for _, ov := range []interface{}{"123", 123, int64(123), float64(123)} {
```

.changeset/fix-not-regex-value-types.md
```diff
@@ -0,0 +1,5 @@
+---
+"posthog-go": patch
+---
+
+Fix `not_regex` local flag evaluation erroring on non-string/int property values. The `not_regex` operator used a manual string/int type switch and returned an error for other types, most notably `float64`, which is what JSON numbers deserialize to, so it failed on a numeric property value even though `regex` handled it. `not_regex` now coerces both sides with `valueToString`, mirroring `regex`.
```
haacked: `nit`: This text lands verbatim in [`CHANGELOG.md`](https://github.com/PostHog/posthog-go/blob/f8425e91b7430a818fd726fc0fc9e0b30749d566/CHANGELOG.md), where it would be the first entry naming an unexported helper, while published entries reference operators or exported identifiers (`is_set`, `icontains`, `FlagKeys`) and describe what the SDK does now rather than what the code used to do.

Try: "Evaluate `not_regex` against all property value types instead of erroring on values that are not a string or int, matching `regex`."

dustinbyrne:
also, your commits will need to be signed in order to be eligible for merge

eeshsaxena:
Good catch, thanks. Fixed: a `nil` property value now returns no-match for `not_regex` rather than being coerced to Go's `<nil>` string. That matches the server and posthog-python, which only allow `None` for `is_not` (`NONE_VALUES_ALLOWED_OPERATORS = ["is_not"]`) and otherwise treat a null property as no-match. Added a regression test (including a `^<nil>$` pattern to pin the exact case you described) and noted it in the changeset.

On signing: understood, I'll get the commits signed so it's merge-eligible.

featureflags.go
marandaneto: Fixed in fa1d6b4. Reproduced with `go test . -run '^TestMatchPropertyNotRegexHandlesAllValueTypes$' -count=1`: the backend-parity case for `nil` with `^1` returned false before the change. Explicit `nil` is now matched as `"null"`, so `^1` yields true and `^null$` yields false; the focused test and `go test ./...` pass. All three PR commits are now signed and verified.

marandaneto (approved):

marandaneto:
blocking: Preserve evaluation-service semantics for null — featureflags.go:1222

 The feature-flags evaluation service represents an explicit JSON null as "null" during regex matching, so not_regex "^1"
  should be true while not_regex "^null$" should be false. This unconditional return makes every not_regex condition
 false for null properties; previously this unsupported value produced an error and triggered authoritative remote
 fallback, whereas the new definitive result can incorrectly disable locally evaluated flags. Reproduction: reproduced —
 go test . -run '^TestReviewNotRegexNullUsesEvaluationServiceSemantics$' -count=1 fails on the reviewed head because
 pattern "^1" returns false instead of true.

i fixed this issue and pushed changes with signed commits so its ready to be merged

PR #310, opened 2026-09-03
feat(distributed-cache): Add distributed cache support

avasenin:

Our tech stack is ruby/go and we would like to use https://posthog.com/docs/feature-flags/local-evaluation/distributed-environments

This PR adds support of this feature for go sdk.

Run tests full suite, please i used example to test out that its working as expected, fetch flags definitions and re-use it and next fetch if needed

## :pencil: Checklist

- [x] I run examples manually with keys and make sure it works as it should

### If releasing new changes

## 🤖 Agent context

- Used existing 3 PRs in ruby/js/python as reference
- Decided to not bring Redis dependency and use file cache in examples to not increase footprint


flag_definition_cache.go
```diff
@@ -0,0 +1,31 @@
+package posthog
+
+import "context"
+
+// FlagDefinitionCacheData is the set of local evaluation data for a project.
+type FlagDefinitionCacheData struct {
```
dustinbyrne, line 6: Since this is a new exported Go struct, could we prevent external unkeyed literals now—for example, with a private zero-sized field (`_ struct{}`)? Otherwise, adding future definition metadata can break consumers using positional `FlagDefinitionCacheData{...}` literals. All current call sites already use keyed literals, so this is inexpensive to establish before release.

flag_definition_cache.go
```diff
@@ -0,0 +1,31 @@
...
+	Flags            []FeatureFlag            `json:"flags"`
+	GroupTypeMapping map[string]string        `json:"group_type_mapping"`
+	Cohorts          map[string]PropertyGroup `json:"cohorts"`
```
dustinbyrne: Could we include `MinimalFlagCalledEvents bool` in this payload and round-trip it through API loads, cache loads, and 304 republishes? The current cache path applies `false`, so a leader can honor the server’s `minimal_flag_called_events` gate while followers emit full `$feature_flag_called` events. Node and Python preserve this value, and the feature-flag-called-tracker spec requires it to survive shared-cache round trips. Since this is an exported struct, adding the field now also avoids changing its public shape later.

featureflags.go
```diff
@@ -512,10 +517,101 @@ func (poller *FeatureFlagsPoller) run() {
...
+		poller.Logger.Errorf("[FEATURE FLAGS] Cache provider GetFlagDefinitions failed: %s", err)
+		return false
+	}
+	if data == nil {
+		return false
+	}
+
+	// The ETag is dropped: a later conditional request must not be answered with 304
+	// against an ETag whose payload this instance no longer holds.
+	poller.applyFlagDefinitions(*data, "", false)
```
dustinbyrne: Could we validate provider data before applying it and retain the current definitions when validation fails? A provider deserializing JSON `{}` returns a non-nil `FlagDefinitionCacheData`, after which this path replaces valid warm definitions with an empty set. A malformed multivariate definition with a missing rollout percentage can also panic during preprocessing. Both are reachable through the file/JSON provider pattern in this PR, and sdk-specs requires malformed provider data not to crash or erase last-known-good state. This can remain scoped to the new cache boundary; an explicitly present empty `Flags` slice should remain a valid cache hit.

featureflags.go
```diff
@@ -2322,6 +2417,7 @@ func (poller *FeatureFlagsPoller) ForceReload() {
 
 func (poller *FeatureFlagsPoller) shutdownPoller() {
 	close(poller.shutdown)
+	poller.shutdownCacheProvider()
```
dustinbyrne: Could we wait for the poll loop and any in-flight provider operation to finish before calling `Shutdown`, and bound cleanup using the caller’s `CloseWithContext` deadline? Closing `poller.shutdown` only signals the loop; an active `ShouldFetchFlagDefinitions` can resume after `Shutdown` completes and call the provider again. I also reproduced a 20 ms `CloseWithContext` remaining blocked because provider cleanup received this separate 30-second deadline. I’m only asking that the newly added provider lifecycle be ordered and bounded correctly, not for a broader poller rewrite.

config.go
```diff
@@ -126,6 +126,12 @@ type Config struct {
 	// polling delay. When set, it overrides DefaultFeatureFlagsPollingInterval.
 	NextFeatureFlagsPollingTick func() time.Duration
 
+	// FlagDefinitionCacheProvider shares local evaluation flag definitions with other
+	// SDK instances through an external cache. Only used when SecretKey is configured.
+	//
+	// EXPERIMENTAL: this API may change in a minor version bump.
+	FlagDefinitionCacheProvider FlagDefinitionCacheProvider
```
dustinbyrne, line 135: Since this is a new feature, could you also add a minor `.changeset/*.md` entry? `RELEASING.md` has instructions for creating one.

examples/flag_definition_cache.go
```diff
@@ -0,0 +1,209 @@
...
+		return
+	}
+	defer os.RemoveAll(dir)
+
+	fmt.Printf("Two instances sharing the cache in %s\n\n", dir)
+
+	clients := make([]posthog.Client, 0, 2)
+	caches := make([]*FileFlagCache, 0, 2)
+
+	for i := 0; i < 4; i++ {
```
dustinbyrne: The function comment, output, slice capacities, and later “both instances” message all describe two clients, but this loop creates four. Could we change the bound to `2`, or update the rest of the example consistently if four is intentional?

dustinbyrne (commented):
hey @avasenin, thanks for putting this together! i've used an agent to compare this to other implementations and have some findings below. i'm around if you want any of these taken off your plate. otherwise i'll leave it to you!

avasenin:
hey @dustinbyrne . i think i finished with all comments. let me know what do u think

featureflags.go
```diff
@@ -512,10 +519,121 @@ func (poller *FeatureFlagsPoller) run() {
...
+// fetchNewFeatureFlags refreshes the local feature flag definitions used for local
+// evaluation. These should not be confused with the feature flags fetched from the
+// flags API.
 func (poller *FeatureFlagsPoller) fetchNewFeatureFlags() {
+	if poller.cacheProvider == nil {
+		poller.fetchFlagDefinitions(false)
+		return
+	}
+
+	shouldFetch, err := poller.cacheProvider.ShouldFetchFlagDefinitions(context.Background())
```
haacked: `blocking`: Client.Close() hangs forever when a cache provider call blocks on the default configuration. The three provider calls in the poll cycle use context.Background() (featureflags.go:531, [featureflags.go:556](https://github.com/PostHog/posthog-go/blob/72cd91d544f419709f7c096b518ca58dab048ec7/featureflags.go#L556), [featureflags.go:618](https://github.com/PostHog/posthog-go/blob/72cd91d544f419709f7c096b518ca58dab048ec7/featureflags.go#L618)), so a provider that respects its context has no deadline to honor. Shutdown waits on these calls to finish.

Here's how it deadlocks: run() closes shutdownDone only when it returns ([featureflags.go:501](https://github.com/PostHog/posthog-go/blob/72cd91d544f419709f7c096b518ca58dab048ec7/featureflags.go#L501)), and it cannot return while fetchNewFeatureFlags is inside a provider call. shutdownPoller waits on shutdownDone or ctx.Done() ([featureflags.go:2448-2452](https://github.com/PostHog/posthog-go/blob/72cd91d544f419709f7c096b518ca58dab048ec7/featureflags.go#L2448-L2452)). In loop(), the deferred shutdownPoller(c.ctx) runs before the deferred close(c.shutdown) because defers run last-registered-first ([posthog.go:1726-1729](https://github.com/PostHog/posthog-go/blob/72cd91d544f419709f7c096b518ca58dab048ec7/posthog.go#L1726-L1729)). ShutdownTimeout defaults to 0 ([config.go:440](https://github.com/PostHog/posthog-go/blob/72cd91d544f419709f7c096b518ca58dab048ec7/config.go#L440)), so Close() calls CloseWithContext(context.Background()) ([posthog.go:1418-1420](https://github.com/PostHog/posthog-go/blob/72cd91d544f419709f7c096b518ca58dab048ec7/posthog.go#L1418-L1420)). A background context's Done() channel is nil, so c.cancel() at [posthog.go:1450](https://github.com/PostHog/posthog-go/blob/72cd91d544f419709f7c096b518ca58dab048ec7/posthog.go#L1450) is never reached and c.ctx is never done. Close() then blocks on <-c.shutdown with nothing left to unblock it.

A Redis-backed provider with a blocked connection stops the process from exiting. The SDK already bounds its HTTP request to /flags/definitions with poller.flagTimeout ([featureflags.go:2406-2410](https://github.com/PostHog/posthog-go/blob/72cd91d544f419709f7c096b518ca58dab048ec7/featureflags.go#L2406-L2410)). Bounding the provider calls the same way caps the poll cycle and guarantees Close() terminates. Cancel right after each call rather than deferring, since the three calls live in three different functions and a deferred cancel in fetchNewFeatureFlags would keep the first timer running across the HTTP fetch and the publish too.

ctx, cancel := context.WithTimeout(context.Background(), poller.flagTimeout)
shouldFetch, err := poller.cacheProvider.ShouldFetchFlagDefinitions(ctx)
cancel()

apply the same pattern at featureflags.go:556 and featureflags.go:618.

featureflags.go
```diff
@@ -2320,8 +2439,19 @@ func (poller *FeatureFlagsPoller) ForceReload() {
 	poller.forceReload <- true
 }
 
-func (poller *FeatureFlagsPoller) shutdownPoller() {
+// shutdownPoller stops the polling loop and then releases the cache provider, so
+// that no provider call is in flight when Shutdown runs. It gives up waiting when
+// ctx is done, bounding cleanup by the deadline the caller passed to Close.
+func (poller *FeatureFlagsPoller) shutdownPoller(ctx context.Context) {
```
haacked: `blocking`: Close() now returns a shutdown timeout error to a client that flushed every event successfully when the process times out waiting for the polling loop to exit. On main, shutdownPoller() closed the shutdown channel and returned immediately. This PR makes it wait for the polling goroutine to exit (featureflags.go:2445-2455), and that wait sits inside loop()'s deferred chain ahead of close(c.shutdown), because deferred calls run last-registered-first ([posthog.go:1725-1728](https://github.com/PostHog/posthog-go/blob/72cd91d544f419709f7c096b518ca58dab048ec7/posthog.go#L1725-L1728)). CloseWithContext is parked on c.shutdown ([posthog.go:1444-1455](https://github.com/PostHog/posthog-go/blob/72cd91d544f419709f7c096b518ca58dab048ec7/posthog.go#L1444-L1455)), so it cannot return until the poll loop is done. The poll loop only checks for shutdown between requests, so an in-flight /flags/definitions call runs to completion, bounded by FeatureFlagRequestTimeout, 3 seconds by default ([config.go:256](https://github.com/PostHog/posthog-go/blob/72cd91d544f419709f7c096b518ca58dab048ec7/config.go#L256)).

The wait also runs after the event-batch drain, so it eats whatever is left of ShutdownTimeout once the events are already out the door. Reproduced empirically: a client with SecretKey set, no cache provider, ShutdownTimeout: 300ms, and a definitions endpoint that takes 2s got a "shutdown timeout: context deadline exceeded" error back from Close() after 302ms, where on main it gets nil. On the default ShutdownTimeout of 0 there is no error, but Close() blocks for the rest of the fetch, because CloseWithContext(context.Background()) hands shutdownPoller a context that is never done.

The doc comment on the new wait says it protects against provider calls being in flight when Shutdown runs. When cacheProvider is nil there is no provider call to protect. Returning early in that case brought Close() back to nil in 21µs in testing and left the whole new cache suite green.

TestShutdownTimeout_AbortsAfterTimeout ([batching_test.go:427](https://github.com/PostHog/posthog-go/blob/72cd91d544f419709f7c096b518ca58dab048ec7/batching_test.go#L427)) sets no SecretKey and so never builds a poller, so nothing in the suite catches this today.

func (poller *FeatureFlagsPoller) shutdownPoller(ctx context.Context) {
	close(poller.shutdown)

	if poller.cacheProvider == nil {
		return
	}

	select {
	case <-poller.shutdownDone:
	case <-ctx.Done():
		poller.Logger.Warnf("[FEATURE FLAGS] Polling loop did not stop before the shutdown deadline: %s", ctx.Err())
	}

	poller.shutdownCacheProvider(ctx)
}

featureflags.go
```diff
@@ -512,10 +519,121 @@ func (poller *FeatureFlagsPoller) run() {
...
+		if flag.Filters.Multivariate == nil {
+			continue
+		}
+		for _, variant := range flag.Filters.Multivariate.Variants {
+			if variant.RolloutPercentage == nil {
+				return fmt.Errorf("flag %q has variant %q without a rollout percentage", flag.Key, variant.Key)
+			}
+		}
+	}
+	return nil
```
haacked: `blocking`: A follower that loads a cache entry with cyclic references crashes with fatal error: stack overflow on the first evaluation touching that flag. Go treats stack overflow as a runtime fatal error that recover() cannot catch, so the host application has no way to survive it. An instance that restarts and reloads the same poisoned entry crashes again.

validateFlagDefinitions ([featureflags.go:577-593](https://github.com/PostHog/posthog-go/blob/72cd91d544f419709f7c096b518ca58dab048ec7/featureflags.go#L577-L593)) only checks that flags is present and that every multivariate variant has a rollout percentage. It does not check references. A cohort whose values hold a property of type cohort pointing back at itself makes matchCohort and matchParsedPropertyGroup call each other without end ([feature_flag_property_group.go:10](https://github.com/PostHog/posthog-go/blob/72cd91d544f419709f7c096b518ca58dab048ec7/feature_flag_property_group.go#L10), :65). Two flags whose dependency_chain entries name each other make evaluateFlagDependency recurse into matchFeatureFlagProperties without end ([featureflags.go:402](https://github.com/PostHog/posthog-go/blob/72cd91d544f419709f7c096b518ca58dab048ec7/featureflags.go#L402)), because a chain entry is only written to evaluationCache after its own evaluation returns.

Both poisoned payloads ran against this branch: validateFlagDefinitions accepted each one, and the first flag evaluation touching either crashed the process with fatal error: stack overflow.

The attacker is whoever can write to the shared cache backing store: another service on the same Redis, an unauthenticated internal Redis, or one compromised instance in the fleet. That same party can hold the leader lock, since it lives in the same store, and keep every other instance reading the poisoned entry. This is not the same as flipping a flag value: flag values stay bounded by what the application does with them, while an uncatchable crash takes the fleet down and cannot be contained by the caller.

One trap before writing the fix: only the dependency_chain drives recursion, so a cycle check must not follow key. Two flags that name each other by key with an empty dependency_chain are what the API sends for a cycle it already caught, and TestFlagDependenciesCircularDependency covers that shape; a key-following check would reject it and strand every follower on that project.

A depth-limit or visited-set guard placed directly in matchCohort and evaluateFlagDependency would close this gap without relying on validateFlagDefinitions alone, and would also cover the pre-existing API path, which shares the same recursive functions and has never had this guard.

Reject cyclic references in validateFlagDefinitions before the data reaches evaluation: a DFS over cohort references and dependency chains, keying off dependency_chain rather than key. Alternatively (or additionally), add a depth-limit or visited-set guard directly in matchCohort and evaluateFlagDependency, which also covers the pre-existing API path.

flag_definition_cache_test.go
```diff
@@ -0,0 +1,746 @@
...
+	require.Equal(t, 2, requests())
+
+	_, _, _, published := provider.calls()
+	require.Len(t, published, 2)
+	require.Len(t, published[1].Flags, 2, "the 304 republishes the definitions in memory")
+	require.Contains(t, published[1].Cohorts, "1")
+	require.True(t, published[1].MinimalFlagCalledEvents, "the 304 republishes the gate too")
+}
+
+func TestFlagDefinitionCacheNotModifiedDoesNotPublishForFollowers(t *testing.T) {
```
haacked: `suggestion`: TestFlagDefinitionCacheNotModifiedDoesNotPublishForFollowers at flag_definition_cache_test.go:463-479 never reaches a 304. The poller starts with no state, so the fetch carries no If-None-Match header ([featureflags.go:2402](https://github.com/PostHog/posthog-go/blob/72cd91d544f419709f7c096b518ca58dab048ec7/featureflags.go#L2402)), the handler's not-modified branch is skipped, and the server returns the full 200 body. A coverage run confirms it: the 304 block at [featureflags.go:656-682](https://github.com/PostHog/posthog-go/blob/72cd91d544f419709f7c096b518ca58dab048ec7/featureflags.go#L656-L682) has zero hits under this test. What the test actually asserts, that a follower's fallback fetch does not publish, is already asserted by TestFlagDefinitionCacheMissWithoutDefinitionsFetchesAnyway.

A follower cannot reach a 304 at all, because it only calls the API when poller.state.Load() is nil ([featureflags.go:542-544](https://github.com/PostHog/posthog-go/blob/72cd91d544f419709f7c096b518ca58dab048ec7/featureflags.go#L542-L544)) and therefore never holds an ETag to send. Deleting the test is cleaner than trying to drive one, since the name currently reads as coverage for a case that cannot happen.

Delete this test.

flag_definition_cache_test.go
```diff
@@ -0,0 +1,746 @@
...
+	require.Contains(t, published[0].Cohorts, "1")
+	require.True(t, published[0].MinimalFlagCalledEvents)
+
+	state := poller.state.Load()
+	require.NotNil(t, state)
+	require.Len(t, state.featureFlags, 2)
+	require.Equal(t, `"etag-1"`, state.flagsEtag)
+}
+
+func TestFlagDefinitionCacheFollowerReadsCacheInsteadOfAPI(t *testing.T) {
```
haacked: `suggestion`: Nothing covers a follower picking up updated definitions from the shared cache on a later poll, which is the feature's steady state. The leader republishes and each follower is supposed to swap in the new flags. TestFlagDefinitionCacheFollowerReadsCacheInsteadOfAPI at flag_definition_cache_test.go:185-213 polls once. The two follower tests that poll twice, TestFlagDefinitionCacheUnusableDefinitionsKeepTheLoadedOnes and TestFlagDefinitionCacheMissKeepsDefinitionsAlreadyLoaded, both feed a miss or malformed data on the second poll, so a warm follower reading valid new content never runs.

If loadFlagDefinitionsFromCache stopped re-applying once state was warm, every follower would serve its first snapshot forever and the suite would stay green. Extending the follower test with a second poll over changed cache content closes it.

flag_definition_cache.go
```diff
@@ -0,0 +1,34 @@
+package posthog
+
+import "context"
+
+// FlagDefinitionCacheData is the set of local evaluation data for a project.
+type FlagDefinitionCacheData struct {
+	_ struct{}
```
haacked: `suggestion`: FlagDefinitionCacheData at flag_definition_cache.go:7 opens with an unexported _ struct{} field with no explanation. No other file in the repo uses this idiom, so a reader meeting it for the first time has no way to tell it is deliberate rather than leftover.

The next person to touch this struct is as likely to delete the field as keep it. Deleting it re-allows unkeyed literals such as FlagDefinitionCacheData{flags, groups, cohorts, true} in user code, and then adding a fifth field to the struct becomes a compile error for everyone who wrote one.

_ struct{} // Forces keyed literals, so adding a field stays non-breaking.

featureflags.go
```diff
@@ -512,10 +519,121 @@ func (poller *FeatureFlagsPoller) run() {
...
+	}
+
+	poller.fetchFlagDefinitions(shouldFetch)
+}
+
+// loadFlagDefinitionsFromCache applies cached definitions and reports whether any
+// were loaded.
+func (poller *FeatureFlagsPoller) loadFlagDefinitionsFromCache() bool {
+	data, err := poller.cacheProvider.GetFlagDefinitions(context.Background())
+	if err != nil {
```
haacked: `suggestion`: When the cache read itself fails (as opposed to returning an empty or missing cache), this instance stops refreshing flag definitions entirely and keeps serving whatever it last loaded for the rest of the process's life. loadFlagDefinitionsFromCache logs the GetFlagDefinitions error and returns false (featureflags.go:557-560), the same value it returns for a legitimately empty cache. fetchNewFeatureFlags then sees non-nil warm state ([featureflags.go:542](https://github.com/PostHog/posthog-go/blob/72cd91d544f419709f7c096b518ca58dab048ec7/featureflags.go#L542)) and returns without ever calling the API.

Ruby and Python both treat a read error as a reason to fetch from the API (setting should_fetch to true in the rescue/except branch), and document this in their provider contract. As long as the Go provider keeps failing its reads, the definitions never change again for the life of the process, with only one log line per poll as a signal.

Distinguish a read error from a legitimate cache miss: return the GetFlagDefinitions error separately from loadFlagDefinitionsFromCache, and treat a non-nil error (rather than just nil state) as a reason to fall back to fetching from the API, matching Ruby and Python.

examples/flag_definition_cache.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+	if err := tmp.Close(); err != nil {
+		return err
+	}
+
+	fmt.Printf("   [%s] published %d flag definitions to the shared cache\n", c.shortID(), len(data.Flags))
+	return os.Rename(tmp.Name(), c.cachePath())
+}
+
+func (c *FileFlagCache) Shutdown(_ context.Context) error {
```
haacked: `suggestion`: Shutdown at examples/flag_definition_cache.go:109 folds an error check and an ownership check into one condition, so the return err on line 112 returns nil in the common case where another instance holds the lock. The behavior is correct, but a reader has to hold both branches in their head to work out that "this isn't my lock" is a success path.

This file is the reference implementation people will copy into a real Redis provider, so it is worth being obvious.

holder, _, err := c.readLock()
if err != nil {
	return err
}
if holder != c.instanceID {
	return nil
}

featureflags.go
```diff
@@ -581,35 +709,26 @@ func (poller *FeatureFlagsPoller) fetchNewFeatureFlags() {
...
-	})
+	// Store new ETag from response (clear if server stops sending)
+	poller.applyFlagDefinitions(data, res.Header.Get("ETag"))
+
+	if publish {
+		poller.publishFlagDefinitions(data)
+	}
 }
 
 // getMinimalFlagCalledEvents reports whether the local-evaluation payload
```
haacked: `question`: When the API responds with a 402 (quota limited), the leader's local reset never reaches followers reading from the shared cache, so they keep serving quota-limited flag definitions indefinitely. fetchFlagDefinitions resets local state to empty flags on a 402 response and returns before reaching the if publish check this PR added (featureflags.go:734), even when this instance is the elected leader (publish == true). Followers call GetFlagDefinitions and apply whatever the cache holds verbatim, so the pre-quota definitions stay live on every follower for as long as the cache entry survives.

Ruby and Python have the same gap in their own 402 handling, so this may be deliberate cross-SDK parity rather than an oversight.

featureflags.go
```diff
@@ -512,10 +519,121 @@ func (poller *FeatureFlagsPoller) run() {
...
+
+		if poller.state.Load() != nil {
+			return
+		}
+
+		// Without definitions local evaluation is impossible, so fetch anyway.
+		poller.Logger.Debugf("[FEATURE FLAGS] No definitions cached or in memory, fetching from the API")
+	}
+
+	poller.fetchFlagDefinitions(shouldFetch)
```
haacked: `question`: Whether only the elected leader ever publishes to the shared cache, versus any instance that successfully fetches from the API, is an undocumented design decision baked into a single boolean parameter. featureflags.go:550 passes shouldFetch (false on this path) as fetchFlagDefinitions's publish argument, so this instance's emergency fetch never calls publishFlagDefinitions. Ruby and Python both publish after any successful fetch, including this emergency one, per their own provider contract docs. Nothing at the call site or in the interface doc says why Go's follower path withholds the publish.

If the cache is empty and multiple followers are in this state, every one of them calls the API independently on every poll instead of the first one populating the cache for the rest, since none of them write what they fetched.

func (poller *FeatureFlagsPoller) publishFlagDefinitions(data FlagDefinitionCacheData) {
	if poller.cacheProvider == nil {
		return
	}
	if err := poller.cacheProvider.OnFlagDefinitionsReceived(context.Background(), data); err != nil {
		poller.Logger.Errorf("[FEATURE FLAGS] Cache provider OnFlagDefinitionsReceived failed: %s", err)
	}
}

featureflags.go
```diff
@@ -581,35 +709,26 @@ func (poller *FeatureFlagsPoller) fetchNewFeatureFlags() {
...
 
 	// Build new groups map
 	groups := map[string]string{}
 	if featureFlagsResponse.GroupTypeMapping != nil {
 		groups = *featureFlagsResponse.GroupTypeMapping
 	}
 
-	// Pre-parse cohort values into typed structs (avoids per-evaluation reconstruction)
-	parsedCohorts := preParseCohortValues(featureFlagsResponse.Cohorts)
+	data := FlagDefinitionCacheData{
```
haacked: `question`: A Ruby or Python instance reading a cache entry written by this Go SDK would silently lose the $feature_flag_id property on $feature_flag_called events. Go publishes data.Flags as a re-marshalled []FeatureFlag (featureflags.go:719-731), and that struct models key, rollout_percentage, active, filters, ensure_experience_continuity, bucketing_identifier, and has_experiment, but not id. The shipped Ruby and Python SDKs store the raw API objects and both read id off the local-evaluation definition to set $feature_flag_id. Neither sibling SDK raises on a missing id key, so the property just stops appearing rather than erroring.

type FeatureFlag struct {
	// Key is the feature flag key.
	Key string `json:"key"`
	// ID is the numeric flag id, used for the $feature_flag_id event property.
	ID *int `json:"id"`
	// RolloutPercentage is the top-level rollout percentage, when configured.
	RolloutPercentage *float64 `json:"rollout_percentage"`
	...
}

flag_definition_cache_test.go
```diff
@@ -0,0 +1,746 @@
...
+	)
+	require.NoError(t, err)
+
+	<-poller.firstFeatureFlagRequestFinished
+	poller.shutdownPoller(context.Background())
+
+	_, _, shutdownCalls, _ := provider.calls()
+	require.Equal(t, 1, shutdownCalls, "shutdownPoller waits for the provider to be released")
+
+	select {
```
haacked: `nit`: The default branch at flag_definition_cache_test.go:530-534 cannot run, because shutdownPoller was handed context.Background() and so only returns once shutdownDone is closed. TestFlagDefinitionCacheShutdownWaitsForAnInFlightProviderCall is what actually proves that ordering, so this select can go.

Remove the unreachable default branch.

featureflags.go
```diff
@@ -512,10 +519,121 @@ func (poller *FeatureFlagsPoller) run() {
...
+		}
+
+		// Without definitions local evaluation is impossible, so fetch anyway.
+		poller.Logger.Debugf("[FEATURE FLAGS] No definitions cached or in memory, fetching from the API")
+	}
+
+	poller.fetchFlagDefinitions(shouldFetch)
+}
+
+// loadFlagDefinitionsFromCache applies cached definitions and reports whether any
```
haacked: `nit`: The doc on loadFlagDefinitionsFromCache at featureflags.go:553 says it reports whether any definitions "were loaded", but TestFlagDefinitionCacheEmptyFlagsIsAHit shows it returns true for a cache entry holding zero flags. "reports whether the cache held usable definitions" matches what the bool actually means.

examples/flag_definition_cache.go
```diff
@@ -0,0 +1,205 @@
...
+
+	if _, err := tmp.Write(encoded); err != nil {
+		tmp.Close()
+		return err
+	}
+	if err := tmp.Close(); err != nil {
+		return err
+	}
+
+	fmt.Printf("   [%s] published %d flag definitions to the shared cache\n", c.shortID(), len(data.Flags))
```
haacked: `nit`: OnFlagDefinitionsReceived at examples/flag_definition_cache.go:105 prints "published N flag definitions" before the os.Rename on line 106 that actually publishes them, so a failed rename prints success and then returns an error. Swap the two lines so the success message only prints after os.Rename succeeds.

Perform the os.Rename first, then print the success message only after it succeeds.

featureflags.go
```diff
@@ -2320,8 +2439,19 @@ func (poller *FeatureFlagsPoller) ForceReload() {
...
+func (poller *FeatureFlagsPoller) shutdownPoller(ctx context.Context) {
 	close(poller.shutdown)
+
+	select {
+	case <-poller.shutdownDone:
+	case <-ctx.Done():
+		poller.Logger.Warnf("[FEATURE FLAGS] Polling loop did not stop before the shutdown deadline: %s", ctx.Err())
+	}
+
+	poller.shutdownCacheProvider(ctx)
```
dustinbyrne, line 2454: **Keep provider cleanup ordered after active polling work**

When `ctx.Done()` wins the select above, this still calls provider `Shutdown` even though the poller has not finished. Provider operations receive `context.Background()`, so client cancellation cannot stop them.

A public-client regression with a context-aware provider and a 20 ms `CloseWithContext` deadline reproduces `Shutdown` overlapping an active `ShouldFetchFlagDefinitions`, followed by `GetFlagDefinitions` after cleanup returns. Depending on the provider, that can use a closed connection or acquire a lock after cleanup already attempted to release it.

Consider keeping the caller's wait bounded while cancelling provider work, preventing further provider calls, and ensuring cleanup runs only after active polling finishes. Add a regression combining an active provider operation with an expiring shutdown deadline; the existing tests cover ordering and timeout separately.

flag_definition_cache.go
```diff
@@ -0,0 +1,34 @@
...
+
+	Flags                   []FeatureFlag            `json:"flags"`
```
dustinbyrne, line 9: **Preserve the complete shared-cache payload rather than the evaluator model**

Using `[]FeatureFlag` makes the cache payload a projection of Go's evaluator model. That type omits fields such as flag `id` and `version`: I reproduced both disappearing when endpoint-shaped JSON is decoded into `FlagDefinitionCacheData` and marshalled again, as the file-cache example does. A different SDK or SDK version reading the same cache may need fields this evaluator does not use.

Consider using `json.RawMessage` at the provider boundary and retaining the original endpoint-shaped definition payload for publication, with typed decoding kept inside the SDK for validation and evaluation. Serializing the existing typed model into `RawMessage` would still lose those fields. Add round-trip coverage preserving `id`, `version`, and unknown metadata, including nested fields.

dustinbyrne (commented):
thanks @avasenin! this is looking great. a couple of additional items as i'm looking through this. again, happy to take anything off your plate if you wish.

lastly - and i missed this on this first look - we require [commit signing](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits) in order to merge to main. can you enable that and rebase/force push to sign your commits?

avasenin:
its still actual and i'm working on it

flag_definition_cache.go
avasenin: i checked other sdks and found that they use kinda hybrid approach. trying to maintain high level structure and put some any data inside. i tried to replicate in recent comments. meaning that i'm using not pure RawMessage but for Flags its []RawMessage and for cohorts its just a Map<string, Raw>

featureflags.go
avasenin: i tried to archieve what you mention. I'm not sure that i like how its end up so you should review it one more time. Its actually become quite complex syncronization logic with cancel context and shutdown channels.

turnipdabeets:
Thanks for this! The **Public API** check is failing because `PropertyMatchingVersion` was added to `FlagDefinitionCacheData`, but the API snapshot wasn't regenerated. To fix it, run:

```sh
make api-update   # same as bin/api-diff --update
```

Then commit the updated `api/public-api.txt`. You can confirm with `make api-diff`, which is what CI runs.

cc @dustinbyrne for another review 🙏

dustinbyrne (approved):
thank you for the contribution @avasenin! it's very much appreciated 🙌

PR #312, opened 2026-09-05
fix(flags): honor versioned local property matching

marandaneto:

Local flag evaluation needs to use the matching rules selected by the definitions response. This follows the [backend change](https://github.com/PostHog/posthog/pull/90694) and [shared SDK contract](https://github.com/PostHog/sdk-specs/pull/59).

- Read `property_matching_version` from definitions. Missing, 1, and unknown versions keep released legacy matching. Exactly 2 uses explicit scalar equality and member equality for nonempty filter arrays. Empty filters keep recursive truthiness, and `is_not` complements `exact` for known properties. For example, version 2 stops matching `false` against `"banana"` and matches `["true", "false"]` against `"true"`.
- Keep the version, flags, dependency index, group mapping, cohorts, and ETag in one atomically replaced definitions snapshot. Each local evaluation pins that snapshot through value/payload APIs, all-flags, `EvaluateFlags`, capture enrichment, and recursive group/cohort/dependency evaluation. A version-only refresh takes effect even when flags are unchanged. A 304 retains the version, including when its ETag changes. Dependency result caches remain per evaluation.
- Fix only the cohort `dependency_chain` decoding seam needed for recursive flag dependencies. JSON decodes chains as `[]any`, while the parser previously accepted only `[]string`. Accept either valid string-only form and leave malformed chains nil for the existing inconclusive path. Other parser behavior is unchanged.
- Preserve server fallback when an active dependency has experience continuity enabled. Resolving a JSON-decoded cohort chain must not bypass the same continuity restriction applied to direct flag evaluation.
- Require server evaluation when a dependency needs group context or is reached from a group-targeted condition. The dependency helper cannot safely reconstruct the referenced flag's person or group properties and bucketing ID. This also covers mixed-targeting flags, nested and negated cohorts, and cached references. A private per-call marker preserves that boundary without disabling ordinary group property-only cohorts. Person-context dependencies on person-only flags still evaluate locally.
- Decode `property_matching_version` in a private wire envelope embedding `FeatureFlagsResponse`. Preserve the released exported struct layout and unkeyed-literal compatibility; the generated public API snapshot now matches main.

This deliberately reduces local coverage for dependencies across group contexts, including mixed-targeting dependencies that also have person conditions. Normal calls can fall back to `/flags`. Local-only value and payload calls stay inconclusive, while bulk, `EvaluateFlags`, and capture paths omit unresolved flags without contacting the server.

Missing-property and ambiguous integral-number safeguards remain unchanged. The existing patch changeset documents the matching and cohort-chain fixes. There is no new disk or shared cache provider, and already-returned evaluation results remain snapshots. This PR does not opt the Go adapter into the [optional harness coverage](https://github.com/PostHog/posthog-sdk-test-harness/pull/53).

**Compatibility:** `FeatureFlagsResponse` retains its released four-field layout. Both keyed and unkeyed consumer literals remain compatible; the matching version is private wire metadata, so the patch changeset is retained.

Added `featureflags_matching_version_test.go` with HTTP definitions-to-local-API tests for missing/1/2/unknown versions, scalar and array matching, negation, null/missing properties, Unicode and canonical values, version-only reloads, 304/error retention, groups, recursive cohorts, flag dependencies, and both capture paths. A deterministic mid-evaluation snapshot replacement verifies that definitions cannot mix across a single evaluation. Dependency-chain tests cover typed, JSON-decoded, and malformed inputs.

Added `featureflags_cohort_continuity_test.go` after independently reproducing the continuity regression. It covers versions 0/1/2/3, enabled and disabled continuity, direct and cohort local-only evaluation, combined value/payload evaluation, raw and pre-parsed cohorts, and cached inconclusive dependencies. The regression failed before the fix and passes afterward. The API snapshot check also failed before regeneration and passes afterward.

Added `featureflags_cohort_group_context_test.go` after independently reproducing both directions of the group-context bug. Its 76 cases cover flag-level and condition-level targeting, versions 0/1/2/3 for referenced group flags and 1/2 for group callers, OR/AND/nested/negated cohorts, indirect dependencies, raw and pre-parsed cohorts, and repeated cached references. Controls verify direct group flags, person-only dependencies, and group property-only cohorts still evaluate locally. Local-only value/payload/bulk/`EvaluateFlags`/capture paths make no remote requests. Both regressions failed before their respective fixes and pass afterward.

Validation for `4bd87debb95d6311abb4c7ab0c835fc00b88fc05`:

```sh
go test -run '^(TestGroupContextPersonFlagDependency|TestCohortFlagDependencyGroupContext)$' -count=1 -v .
go test -race -count=1 -timeout=5m ./...
go vet ./...
bin/api-diff
git diff --check
```

All final checks passed. The full race suite passed all root-module packages, with the root package completing in 47.016s. Changed Go files are formatted. An earlier race run hit an existing shutdown timing assertion, and an intermediate test fixture needed its logger initialized. Both logs were retained and the final full suite passed. The first required committed-branch autoreview found the inverse group-to-person dependency bug. That finding was independently reproduced and fixed in a second new signed commit. The final isolated autoreview against `origin/main` reported no actionable findings for this exact commit. Live-service and cross-SDK harness integration were not run. These are local results, not a claim that CI passed.

### Review follow-up validation (`1c7c75e67f758c56342026cb18e11cd1d0639f28`)

Merged canonical main `8cba50ae08efdfde25b71f34a322b0932fca54ce` without conflicts in `b5994a322672230f7a60221c75fba250a7bca915`. After explicit human approval, restored `FeatureFlagsResponse` compatibility with a private decoding envelope. Before the fix, an external consumer using `posthog.FeatureFlagsResponse{nil, nil, nil, false}` and the retained `TestFeatureFlagsResponseUnkeyedLiteralCompatibility` both failed to compile (too few struct values). Both pass after the fix. The retained test also checks decoding of all four public fields when new metadata is present.

Focused compatibility/version/snapshot/group tests, `go test -race -count=1 -timeout=5m ./...` (root: 47.088s), `go vet ./...`, `go build ./...`, `bin/api-diff`, and `git diff --check` passed. The regenerated API snapshot is identical to main. These are local results; see current checks for CI results.

## :pencil: Checklist

### If releasing new changes

The existing `.changeset/versioned-property-matching.md` patch entry is included. The generator was not rerun. Public struct-literal compatibility is preserved as described above.

## 🤖 Agent context

Pi agents implemented, independently reviewed, and updated this change under human direction using file tools, Git, Go tests, GitHub CLI, and the isolated autoreview helper. The review fixes preserve continuity and group-context fallback boundaries and refresh the generated API snapshot. A private group-context marker was approved after the inverse dependency bug was reproduced. No public API or matching-rule changes were added by these fixes. The human-directed scope includes the narrow JSON dependency-chain fix, not broader parser cleanup. Human review is required before merge.

Session context is local and has no shareable URL. Review, reproduction, and publication evidence were retained outside the repository.

### DIFF
--- a/.changeset/versioned-property-matching.md
+++ b/.changeset/versioned-property-matching.md
@@ -1,0 +1,5 @@
+---
+"posthog-go": patch
+---
+
+Honor the definitions snapshot's `property_matching_version` during local flag evaluation: version 2 uses explicit boolean equality, while missing/1 and unknown versions retain legacy matching. Keep definitions and matching semantics together across reloads, cached 304 responses, group/cohort rules, and flag dependencies. Also preserve JSON-decoded dependency chains in cohort leaves.
--- a/.changeset/versioned-property-matching.md
+++ b/.changeset/versioned-property-matching.md
@@ -2,4 +2,4 @@
 "posthog-go": patch
 ---
 
-Honor the definitions snapshot's `property_matching_version` during local flag evaluation: version 2 uses explicit boolean equality, while missing/1 and unknown versions retain legacy matching. Keep definitions and matching semantics together across reloads, cached 304 responses, group/cohort rules, and flag dependencies. Also preserve JSON-decoded dependency chains in cohort leaves, while requiring server evaluation for dependencies that need group context.
+Honor the definitions snapshot's `property_matching_version` during local flag evaluation: version 2 uses explicit boolean equality, while missing/1 and unknown versions retain legacy matching. Keep definitions and matching semantics together across reloads, cached 304 responses, group/cohort rules, and flag dependencies. Also preserve JSON-decoded dependency chains in cohort leaves, while requiring server evaluation for dependencies that need group context or are reached from group-targeted conditions.
--- a/.changeset/versioned-property-matching.md
+++ b/.changeset/versioned-property-matching.md
@@ -2,4 +2,4 @@
 "posthog-go": patch
 ---
 
-Honor the definitions snapshot's `property_matching_version` during local flag evaluation: version 2 uses explicit boolean equality, while missing/1 and unknown versions retain legacy matching. Keep definitions and matching semantics together across reloads, cached 304 responses, group/cohort rules, and flag dependencies. Also preserve JSON-decoded dependency chains in cohort leaves.
+Honor the definitions snapshot's `property_matching_version` during local flag evaluation: version 2 uses explicit boolean equality, while missing/1 and unknown versions retain legacy matching. Keep definitions and matching semantics together across reloads, cached 304 responses, group/cohort rules, and flag dependencies. Also preserve JSON-decoded dependency chains in cohort leaves, while requiring server evaluation for dependencies that need group context.
--- a/featureflags.go
+++ b/featureflags.go
@@ -67,11 +67,12 @@ func getOrCompileRegex(pattern string) (*regexp.Regexp, error) {
 // flagsState holds the feature flag data that is atomically swapped during updates.
 // This provides lock-free reads for the common path (flag evaluation).
 type flagsState struct {
-	featureFlags []FeatureFlag
-	flagsByKey   map[string]FeatureFlag // pre-built index for O(1) lookup, avoids rebuilding per evaluation
-	cohorts      map[string]PropertyGroup
-	groups       map[string]string
-	flagsEtag    string
+	featureFlags            []FeatureFlag
+	flagsByKey              map[string]FeatureFlag // pre-built index for O(1) lookup, avoids rebuilding per evaluation
+	cohorts                 map[string]PropertyGroup
+	groups                  map[string]string
+	flagsEtag               string
+	propertyMatchingVersion int
 	// minimalFlagCalledEvents is the server-controlled gate for minimal
 	// $feature_flag_called events, cached from the local-evaluation payload.
 	minimalFlagCalledEvents bool
@@ -221,6 +222,9 @@ type FlagVariantMeta struct {
 
 // FeatureFlagsResponse is the wire-format response from the local evaluation endpoint.
 type FeatureFlagsResponse struct {
+	// PropertyMatchingVersion selects exact/is_not semantics; only 2 enables explicit matching.
+	// Missing versions retain legacy matching.
+	PropertyMatchingVersion int `json:"property_matching_version"`
 	// Flags contains feature flag definitions for local evaluation.
 	Flags []FeatureFlag `json:"flags"`
 	// GroupTypeMapping maps group type indexes to group type names.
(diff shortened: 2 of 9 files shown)


.changeset/versioned-property-matching.md
```diff
@@ -0,0 +1,5 @@
+---
+"posthog-go": patch
```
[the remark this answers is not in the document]
marandaneto: Fixed in 1c7c75e67f758c56342026cb18e11cd1d0639f28. Kept the released four-field FeatureFlagsResponse layout and decode property_matching_version only in a private wire envelope, preserving the patch release and existing snapshot semantics. Before the fix, an external consumer with FeatureFlagsResponse{nil, nil, nil, false} and the retained TestFeatureFlagsResponseUnkeyedLiteralCompatibility both failed to compile with “too few values in struct literal”; both pass now. The retained test also verifies all four public fields still decode when version metadata is present. Existing version/reload/304/snapshot/group tests pass, as do go test -race -count=1 -timeout=5m ./..., go vet ./..., go build ./..., and bin/api-diff. The regenerated public API snapshot is identical to main. All current checks on this exact commit passed, including both Go matrices and compliance v0/v1: https://github.com/PostHog/posthog-go/actions/runs/35335553648 and https://github.com/PostHog/posthog-go/actions/runs/35335554242.

featureflags.go
```diff
@@ -335,7 +336,15 @@ func (poller *FeatureFlagsPoller) evaluateFlagDependency(
 	deviceId *string,
 	properties Properties,
 	cohorts map[string]PropertyGroup,
+	groupContext bool,
+	snapshots ...*flagsState,
 ) (bool, error) {
+	state := poller.evaluationState(snapshots)
+	// Group conditions no longer have the person's properties or bucketing ID.
+	// This must precede cache lookup, including for person-targeted dependencies.
+	if groupContext {
```
turnipdabeets: [question] Main evaluated a company flag that depends on another company flag locally and correctly (same group key, same group properties). With this it always goes to `/flags`, or gets dropped under `OnlyEvaluateLocally`. Could we allow the dependency when its aggregation matches the caller's, the way Python and Node do, and add a test for that case? The cross-type fix makes sense to keep.

turnipdabeets (approved):

featureflags.go
marandaneto: Fixed in 6476c8300b5e0ed3973a001dcdc9d0fd760b56c7. Same-aggregation dependencies now reuse the caller’s group key and properties, including group-targeted conditions and nested cohorts. Cross-type dependencies (person/group, different groups, or mixed conditions needing another group) still require server evaluation, checked before cached results.

Reproduced with `go test -count=1 -run "^TestSameGroupFlagDependency$" .`: before the fix, same-company dependencies errored locally, called `/flags/` with fallback enabled, and disappeared from local-only bulk/evaluation/capture results. The test now passes across matching versions 0–3, direct/cohort/nested/transitive dependencies, and matching/nonmatching properties; its rollout fixture distinguishes the group key from the person ID. Added cached/negated cross-group controls and retained the existing person↔group regressions.

Full race tests, vet/build, formatting, module tidy, OTel checks, and public API snapshot pass. All checks passed on this exact SHA, including Go matrices and compliance v0/v1: https://github.com/PostHog/posthog-go/actions/runs/35426033341 and https://github.com/PostHog/posthog-go/actions/runs/35426033738.

PR #314, opened 2026-09-09
fix: omit null custom event properties during serialization

marandaneto:

Custom event properties currently send explicit null object members. This implements the serialization rule in https://github.com/PostHog/sdk-specs/pull/60 for the Go SDK's legacy and v1 event paths.

The event serializer now omits null object members recursively, including objects inside arrays. It preserves null array positions, empty objects and arrays, false, zero, strings, supported numbers, and ordered duplicate JSON members. Existing nullable public inputs and caller-owned values remain unchanged. Cleanup happens after enrichment and before-send processing without restoring privacy-filtered fields.

The shared event serialization boundary covers capture, AI-named ordinary capture, exception custom properties, and identify/group custom properties. It does not change generic `Properties` JSON or introduce a separate AI route. Existing typed flag response nulls and exception metadata keep their field-specific behavior. There is no blanket exemption for `$` or AI properties and no disk queue added.

The correction also preserves the original concrete serialization callback errors. Only the extra wrapper from the private event-property adapter is removed. User marshaler error layers and the existing exception wrapper remain intact. The zero-values fixture now expects null omission, with its nil input and all other assertions preserved.

**Current draft status:** The signed and reviewed HEAD is `28ea011afc52a967565102d20c3576671e919007`. This includes the correction for the stale null fixture and callback error wrapper that caused four test failures at the previous HEAD. All 28 hosted checks passed for this exact SHA, including Test and Race jobs on Go 1.21 and Go 1.x, Public API, builds, vet, formatting, OTel, security scans, and v0/v1 compliance. Unit Tests: https://github.com/PostHog/posthog-go/actions/runs/34338572018. The bounded watch completed successfully on 2026-09-09. These results supersede the four failures on the previous SHA.

- Passed 43 selected top-level Go tests with `-race -count=1` on Go 1.26.4, Darwin/arm64. These include 84 exact-byte ordered-member cases, 8 malformed/out-of-range rejection controls, and 24 concrete error-chain and user-error identity controls across both protocols and capture/identify/group/exception paths.
- Actual SDK enqueue/close and feature flag evaluation paths use mocked transports or numeric-loopback HTTP servers with fake keys. Tests inspect received serializer bytes for nested nulls, array positions, hooks, caller nonmutation, all-null events, typed flag metadata, exception properties, and privacy filtering. Compatibility tests check actual client failure callbacks as well as preparation errors.
- Passed `go vet -mod=readonly .`, changed-file `gofmt` checks, and branch whitespace checks. Reused existing offline caches with module downloads disabled. No dependencies were installed.
- The installed isolated committed-head autoreview passed at `28ea011afc52a967565102d20c3576671e919007` against main `53f54712427369388565577777d13970689395e1`, with mandatory TruffleHog clean and no findings. Main is included without a merge because it has not advanced.

**Draft validation limits:** The full local API snapshot check remains blocked by existing uncached CLI/example dependencies (`godotenv`, `urfave/cli`, and `go-md2man/v2`), as recorded during the previous attempt. It was not retried or repaired here. The full suite, minimum-Go matrix, local-poller/transport-failure flags, retry/compression-specific null payloads, and performance checks were not run locally. Existing goccy rejection of `1e400` remains unchanged. Disk persistence is not applicable. Hosted results are separate from these local limits. Existing compliance harness passes do not establish the missing null-policy fixture coverage.

## :pencil: Checklist

### If releasing new changes

An equivalent `posthog-go: patch` changeset was added manually. The same changeset covers the correction. No package-manager setup or release was run.

## 🤖 Agent context

Implemented and validated with Pi coding assistance, Git, Go, GitHub CLI, and the installed isolated autoreview helper. The human-directed scope was recursive event-property null omission without public API narrowing, array compaction, or duplicate-key loss. Existing goccy token traversal was used instead of a map round-trip so supported duplicate members and numeric values survive. The callback correction removes only the private adapter's added error layer. Private session logs are not published.

This remains a draft for human review, assigned to marandaneto. No release, auto-merge, or approval is requested by the agent.

### DIFF
--- a/.changeset/tidy-otters-listen.md
+++ b/.changeset/tidy-otters-listen.md
@@ -1,0 +1,5 @@
+---
+"posthog-go": patch
+---
+
+Omit null-valued custom event object properties recursively while preserving array positions, supported JSON values, and typed flag and exception metadata.
--- a/capture_v1.go
+++ b/capture_v1.go
@@ -272,7 +272,7 @@ func prepareForSendV1(msg Message, logger Logger) (json.RawMessage, APIMessage, 
 		Properties eventProperties `json:"properties"`
 	}{ev, props})
 	if err != nil {
-		return nil, apiMsg, ev.Uuid, err
+		return nil, apiMsg, ev.Uuid, eventPropertySerializationError(err)
 	}
 	return json.RawMessage(data), apiMsg, ev.Uuid, nil
 }
--- a/capture_v1.go
+++ b/capture_v1.go
@@ -257,7 +257,20 @@ func baseV1Props(isServer bool, disableGeoIP bool) Properties {
 func prepareForSendV1(msg Message, logger Logger) (json.RawMessage, APIMessage, string, error) {
 	apiMsg := msg.APIfy()
 	ev := buildV1Event(msg.apifyEvent(), logger)
-	data, err := json.Marshal(ev)
+	props := flagEventProperties(ev.Event, ev.Properties)
+	if exception, ok := apiMsg.(ExceptionInApi); ok {
+		props.preserve = []string{"$exception_list"}
+		if exception.Properties.ExceptionFingerprint != nil {
+			props.preserve = append(props.preserve, "$exception_fingerprint")
+		}
+		if len(exception.Properties.DebugImages) > 0 {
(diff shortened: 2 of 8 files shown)


marandaneto:
gonna be done server side anyway

PR #320, opened 2026-09-11
feat!: capture v1 pre-release + capture_ai - DO NOT MERGE (v2.0.0)

eli-r-ph:
Makes capture v1 the only capture path and releases it as posthog-go 2.0.0. The legacy `/batch/` pipeline and the `Config.CaptureMode` switch are gone, and the v1 internals take back the unsuffixed names.

Base for the stacked `v1-*` PRs to follow.

## Changes

- **Removed the legacy capture pipeline** — `/batch/` send path, `Config.CaptureMode` and its constants, and the `capturer` strategy interface.
  - `capturer` is deleted rather than kept for a future `CaptureAI()`: posthog-rs added its isolated AI lane with zero traits, varying endpoint/compression/queue/byte-target through a plain config value, which is a different axis than `capturer` abstracts.
- **Ported the v0-only test coverage onto the v1 wire path** rather than deleting it; duplicates of existing v1 tests were dropped.
  - Root-package tests go 1247 → 1208, and every one of the 45 removals is enumerated in the commit message.
- **Renamed `*V1` internals to the now-free plain names**, in a commit with zero logic changes.
  - Route strings, the `capture_v1` harness capability and `PostHog-*` headers keep their `v1` — they are external contract, not internal markers.
- **Collapsed the SDK compliance rig to one adapter and one CI job**; the adapter advertises only `capture_v1`, which deselects the v0 suite.
- **Moved the module to `github.com/posthog/posthog-go/v2`** — Go encodes the major in the import path, so this is required to release 2.0.0, not cosmetic.
- **Decoupled the otel bridge's version from the core.** It previously tagged `otel/v${ROOT_VERSION}`, so a core major would publish an `otel/v2.0.0` that Go refuses to resolve, on every 2.x release.
- **Fixed a terminal-status error losing its cause** — a 400 with an unreadable body now unwraps to the read error instead of arriving as a bare `capture request failed: 400`.

## Behavior changes with no compile-time signal

Documented in `docs/migration-v2.md` and the changeset. The endpoint moves to `/i/v1/analytics/events`, auth moves to `Authorization: Bearer`, results become per-event (so `Callback.Failure` can fire on a 200), 429 is no longer retried, `$lib`/`$lib_version` leave event properties, and caller properties now win over SDK system context.

Proxies, egress allowlists and self-hosted deployments pinned to `/batch/` need updating.

## Verification

- 1230 tests pass across all packages with `-race`, plus the otel module. `TestOpenMachOUniversal` fails here and on `main` — darwin-only, needs an `LC_UUID` local builds lack, never runs in CI.
- Full 94-test smoke suite green against production US: **89 passed, 0 failed**.
- Confirmed in prod ClickHouse that `$lib`/`$lib_version` arrive even though the SDK no longer sends them, proving the `PostHog-Sdk-Info` header path works end to end.

## Follow-ups

- [PostHog/posthog#99499](https://github.com/PostHog/posthog/pull/99499) moves `bin/hobby-installer` to `/v2`; it cannot merge until 2.0.0 is published.
- The Changesets release flow now has two packages — write changesets against `posthog-go-otel` for otel-only changes.


posthog.go
```diff
@@ -1759,22 +1592,20 @@ func (c *client) loop() {
...
-		if msgSize > maxMessageBytes {
-			c.Errorf("message exceeds maximum size (%d > %d)", msgSize, maxMessageBytes)
+		if msgSize > c.MaxEventBytes {
+			c.Errorf("message exceeds maximum size (%d > %d)", msgSize, c.MaxEventBytes)
 			c.notifyFailure([]APIMessage{prepared.msg}, ErrMessageTooBig)
 			return false
 		}
 
-		if batchSize+msgSize > maxBatchBytes && len(batchData) > 0 {
+		if batchSize+msgSize > c.MaxBatchBytes && len(batchData) > 0 {
```
marandaneto, line 1601: blocking: Enforce the independently configured batch byte limit for single events — When MaxBatchBytes is lower than MaxEventBytes (including lowering only MaxBatchBytes), an event exceeding the request limit passes the event-size check. This condition only flushes an existing batch and then appends the oversized event to the empty batch, so callers configuring a proxy body limit still send requests that can receive terminal 413 responses. Handle events that cannot fit the independently configured request limit. Reproduction: reproduced — `go test -mod=readonly -run '^TestReviewMaxBatchBytesBoundsSingleRequest$' -count=3 .` in a disposable exact-head worktree failed all three runs because a 1,000-byte MaxBatchBytes permitted a 2,313-byte request.

marandaneto (commented):
Automated advisory code review. Findings validated against the reviewed revision with temporary regression tests.

marandaneto:
draft until we merge the rest of the prs

PR #322, opened 2026-09-12
fix: order queue sends against shutdown, and surface silent event loss

eli-r-ph:
Two pre-existing bugs found while building the CaptureAI lane, both reproducible on `main` and so present in the published 1.x SDK. Stacked on #321; the CaptureAI PR stacks on this.

## Changes

- **Fixed a data race between `Enqueue` and `Close`.** The batch loop closes the message queue on shutdown, so a concurrent `Enqueue` was a send on a closed channel — a data race by the memory model, not merely the panic the SDK recovered from. Any application whose graceful shutdown overlaps an enqueue could trip the race detector in its own test suite.
  - Sends now hold a read lock that `Close` takes for write when it marks the client closed. A send either completes before the queue closes or returns `ErrClosed`, which is what callers already saw.
  - `BenchmarkConcurrentEnqueue` at 1/10/100/500/1000 goroutines is unchanged: 1105–1298 ns/op with the fix against 1097–1243 without, overlapping at every count. The lock disappears into the ~1.2 µs `Enqueue` already spends on enrichment and serialization.

- **A capture failure that reaches no `Callback` now logs one aggregate line per batch instead of nothing.** A client with a bad API key and no `Callback` previously saw complete silence. This covers terminal responses (`400`, `401`, `402`, `413`, `415`), shutdown during retry backoff, and events the server drops inside a `200`.
  - The line carries a count and a cause, never per-event detail: per-event logging scales with event volume rather than request volume, and payloads may carry sensitive content. A registered `Callback` suppresses it, since it already reports every failure.

## Verification

The race is proven pre-existing, not introduced by the v1 branch — the same probe reports a race on `main` @ `24cd918`, which is published v1.25.1.

Each new test was verified to fail when the thing it guards is broken: reverting the lock trips the detector, removing the warn fails both loss tests, dropping the `Callback` gate fails both suppression tests, and reverting only the terminal-non-2xx call site fails only that subtest.

Full suite and `-race` are green apart from `TestOpenMachOUniversal`, which fails identically on `main` (darwin-only, needs an `LC_UUID` local builds lack, never runs in CI).


marandaneto (approved):

.changeset/shutdown-race-and-silent-loss.md
```diff
@@ -0,0 +1,11 @@
...
+"posthog-go": patch
+---
+
+Fix a data race between `Enqueue` and `Close`, and stop losing events silently.
+
+- `Enqueue` racing `Close` was a send on a closed channel — a data race, not merely the panic the SDK recovered from. Any application whose graceful shutdown overlaps an enqueue could trip the race detector in its own test suite. Sends are now ordered against shutdown, and still return `ErrClosed` as before.
+
+- A capture failure that reaches no `Callback` now logs one aggregate line per batch instead of nothing. This covers terminal responses (`400`, `401`, `402`, `413`, `415`) and events the server drops inside a `200`. Previously a client with a bad API key and no `Callback` configured saw complete silence.
+
+  The line carries a count and a cause, never per-event detail: per-event logging scales with event volume rather than request volume, and payloads may carry sensitive content. Registering a `Callback` suppresses it, since the callback already reports every failure.
```
marandaneto: this should be one liner

PR #323, opened 2026-09-12
feat: add EnqueueAI on a dedicated capture-ai lane

eli-r-ph:
Adds `Client.EnqueueAI`, sending to PostHog's dedicated AI capture endpoint (`/i/v1/ai/events`) on an isolated second pipeline. Stacked on #322.

## Changes

- **`EnqueueAI` runs on its own lane** — own queue, batching and retry state, started lazily on first use so clients that never send AI events pay nothing. The AI endpoint accepts events up to 8MiB against the analytics endpoint's much smaller limit, so a shared pipeline would either reject large AI events or loosen the analytics limit for everything.

- **Routing is by method, never by event name.** `Enqueue` does not reroute an `$ai_`-prefixed event. Customers send custom `$ai_*` names as ordinary analytics events at volume, and the server owns misrouting.

- **`EnqueueAIWithContext(ctx, client, msg)`** is the request-context-aware form, matching `EnqueueWithContext`. It takes the full `Client` because only a real client has an AI lane — an AI event must never silently fall back to the analytics endpoint.

- **New config:** `Config.CaptureAICompression` and `Config.CaptureAIMaxQueueSize` (default `DefaultCaptureAIMaxQueueSize`, 1000). The lane's per-event (8MiB) and per-batch (5MiB) byte limits track the server's and stay internal constants, matching posthog-rs, posthog-python and posthog-node.

- **`CaptureEventError` and `CaptureRequestError` gained `Endpoint`**, so one `Callback` can tell the two lanes apart.

- **Breaking for implementers only:** `Client` gained `EnqueueAI`, so a hand-written mock of the interface must add it. Callers are unaffected.

## Shutdown ordering

`Close` marks the client closed *before* it reads the AI lane, and the lane re-checks after starting. A caller that observes "not closed" either started the lane in time for `Close` to see and drain it, or observes "closed" on the re-check and declines to use it — so a lane can never be started after `Close` has already looked. Both lanes share one shutdown deadline.

This is the hazard that bit posthog-rs twice in review, so it is pinned by a repeated-race test. A probe confirms the window is genuinely exercised rather than always resolving one way: the lane is started concurrently with `Close` in 93 of 200 iterations.

## Structure

The first commit moves the existing pipeline onto an explicit `lane` value and is behaviour-neutral — 421 test functions before and after, with an empty name diff verified against the base in a worktree. The second adds the AI lane itself.

## Verification

Full suite and `-race` are green apart from `TestOpenMachOUniversal`, which fails identically on `main` (darwin-only, needs an `LC_UUID` local builds lack, never runs in CI).


capture_lane.go
```diff
@@ -0,0 +1,122 @@
...
+// declines to use it. Without the re-check a lane could be started after Close
+// had already looked, and nothing would wait for it.
+func (c *client) aiLane() *lane {
+	if c.closed.Load() {
+		return nil
+	}
+	c.aiOnce.Do(func() {
+		l := newLane(aiLaneConfig(c.Config), c.MaxEnqueuedRequests)
+		c.ai.Store(l)
+		go c.loop(l)
```
marandaneto, line 113: blocking: Shut down the shared feature-flag poller only once — With SecretKey (or PersonalApiKey) configured, EnqueueAI starts a second c.loop, but each loop still defers c.featureFlagsPoller.shutdownPoller(). Ordinary Close() therefore makes both loops close the same poller channel, causing an unrecovered background-goroutine panic that terminates the application. Keep poller lifetime client-owned rather than independently owned by each lane. Reproduction: reproduced — `go test -count=1 -timeout=30s -run '^TestReview323CloseAIWithSecretKey$' .` in a disposable exact-head worktree failed with `panic: close of closed channel` in FeatureFlagsPoller.shutdownPoller.

config.go
```diff
@@ -154,6 +154,17 @@ type Config struct {
 	// zero, it defaults to DefaultMaxBatchBytes.
 	MaxBatchBytes int
 
+	// CaptureAICompression selects the compression mode for EnqueueAI request
+	// bodies, independently of Compression. If zero, AI bodies are sent
+	// uncompressed. CompressionZstd is a good choice: AI events are large JSON.
+	CaptureAICompression CompressionMode
```
marandaneto, line 161: blocking: Validate the AI compression mode during construction — CaptureAICompression is consumed by upload, but Config.Validate() still checks only Compression. An unsupported value such as CompressionMode(255) therefore constructs a client successfully; every AI upload then fails locally in compressBody and enters the retry/drop path without making an HTTP request. Validate this setting alongside analytics compression so a permanent configuration error is surfaced before events are accepted. Reproduction: reproduced — `go test -count=1 -timeout=30s -run '^TestReview323RejectInvalidAICompression$' .` in a disposable exact-head worktree failed because NewWithConfig returned no error for the unsupported AI codec.

marandaneto (commented):
Automated advisory code review. Findings validated against the reviewed revision with temporary regression tests.

marandaneto (approved):
bots still has some comments otherwise lgtm

.changeset/capture-ai-lane.md
```diff
@@ -0,0 +1,19 @@
...
+
+- `EnqueueAIWithContext(ctx, client, msg)` is the request-context-aware form, matching `EnqueueWithContext`.
+
+- `Config.CaptureAICompression`, `Config.CaptureAIMaxQueueSize` (default `DefaultCaptureAIMaxQueueSize`, 1000) and `Config.CaptureAIBatchUploadTimeout` (default `DefaultCaptureAIBatchUploadTimeout`, 30s) configure the AI lane independently. Its per-event (8MiB) and per-batch (5MiB) byte limits track the server's and are not configurable.
+
+  The AI lane gets a longer upload timeout than the analytics lane's `BatchUploadTimeout` because its batches are an order of magnitude larger. The two lanes share one HTTP transport, so the connection pool is still shared.
+
+- `CaptureEventError` and `CaptureRequestError` gained `Endpoint`, so a single `Callback` can tell the two lanes apart.
+
+- Breaking for implementers only: `Client` gained `EnqueueAI`, so a hand-written mock of the interface must add it. Callers are unaffected.
```
marandaneto: this should be a one liner

posthog.go
```diff
@@ -1490,48 +1540,48 @@ func (c *client) processBatch() {
...
 	}
 
 	// Blocking send with timeout - allows in-flight requests to complete during latency spikes
 	timer := time.NewTimer(c.BatchSubmitTimeout)
 	select {
-	case c.batches <- batch:
+	case l.batches <- batch:
 		timer.Stop()
-		go c.processBatch()
+		go c.processBatch(l)
```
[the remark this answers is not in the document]
eli-r-ph: this is std behavior in posthog-go on main and in std capture atm, not in scope for this change

Issue #229, opened 2026-06-18
Add public reusable Flush API


marandaneto:
## Summary
Add a public reusable `Flush()` API to the Go SDK.

## Spec
- `sdk-specs/openspec/specs/flush/spec.md`

## Current evidence
The SDK audit found `Client` exposes `Close()` / `CloseWithContext()`, but no reusable immediate flush API. `Close()` drains during shutdown, but it also closes/disables the client.

## Expected behavior
- Expose a public `Flush()` method that immediately sends queued events without closing the client.
- Ensure the client remains usable after a successful flush.
- Respect errors/timeouts according to existing Go SDK conventions.
- Add tests that queued events are delivered and later events can still be enqueued.

cristiangirlea:
@PostHog/team-client-libraries

Following up from the review on #324. ioannisj asked for the API shape to be agreed here before the PR moves on, so here is what I'm proposing and why.

**What I need**

A way to push queued events out without closing the client. `Close()` drains, but then the client is gone, which doesn't work for request handlers in serverless, CLI tools that checkpoint partway through, or tests that want deterministic delivery. `BeforeSend` doesn't help because it sits on the enqueue side, not delivery.

**Proposed shape**

```go
// on Client, next to Close / CloseWithContext
Flush() error
FlushWithContext(context.Context) error
```

`Flush` is just `FlushWithContext(context.Background())`. Cancelling the context stops the caller's wait, not delivery, and the client stays usable afterwards. Calling either after `Close` returns `ErrClosed`.

I went with the pair rather than a single ctx-taking method because that's how `Close` / `CloseWithContext` already work, and `flush` is the name every other SDK uses. The noop client gets both and returns nil.

**Behaviour**

I checked this against the flush spec in sdk-specs and it lines up:

- Empty queue returns immediately, nothing goes over the wire.
- Whatever is queued at call time gets sent now, regardless of batch size or interval.
- The call returns once every covered batch has either finished an attempt or gone into retry backoff. On a 5xx, 429 or network error the transport keeps the batch on its normal retry schedule. Flush doesn't skip the backoff and doesn't sit through it either.
- The only errors it returns are `ErrClosed` and `ctx.Err()`. Delivery failures still go through `Callback.Failure`, so a nil return doesn't mean ingested.
- Concurrent flushes go through the batch loop and each one waits on what was queued at its own barrier. Events enqueued mid-flush might land in this cycle or the next one.

[...]

PR #324, opened 2026-09-14
feat: add reusable client flush

cristiangirlea:

Related to #229.

Adds `Flush()` and `FlushWithContext(ctx)` to send queued events without closing the client. Cancelling the context stops the caller's wait, not delivery or future use of the client.

A retryable failure ends the current flush cycle. The transport keeps the event for its scheduled retry, including the existing backoff and retry budget. Flush waits for active attempts but does not skip retry backoff. Terminal failures still use `Callback.Failure`, so a nil return is not a delivery guarantee.

This is a draft for API review. Adding methods to `Client` affects custom implementations and mocks, so the changeset proposes a major release. I'd appreciate guidance on whether this should target the current API or the upcoming v2/capture-lane work. A blocking Flush also cannot be called from a delivery callback.

Tested on Go 1.21.13 in an offline Docker sandbox, based on main at `07b16ce`:

- Eleven new top-level tests cover reuse, empty queues, cancellation, concurrent calls, in-flight delivery, retries and terminal callbacks across both capture modes.
- The 503 regression failed against the earlier candidate. It now checks that Flush returns while the event is retained, then verifies eventual delivery and client reuse.
- Twenty race-enabled runs of the Flush tests passed.
- `make build` passed, including vet and the full race-enabled suite.
- A separate consumer module passed two enqueue/flush cycles against a local HTTP server.
- The API snapshot check passed with offline dependency-lookup warnings.

Current-Go, nested OTel and the official compliance harness have not been run. These are local results, not hosted CI or production-backend results.

## :pencil: Checklist

### If releasing new changes

The major changeset was written manually.

## 🤖 Agent context

Codex helped implement and test this using GitHub tools and Docker. The change reuses the batch loop and transport retry machinery rather than introducing a separate retry queue.


posthog.go
```diff
@@ -66,6 +66,21 @@ type EnqueueClient interface {
 type Client interface {
 	EnqueueClient
 
+	// Flush sends queued messages and waits for the current delivery cycle without
```
ioannisj, line 69: Thanks for picking this up. 

Before we go further, I think we need to agree on the API shape on #229, since this adds `Flush` and `FlushWithContext` to the exported `Client` interface and ships as a major bump.

The "Public API changes" section in `CONTRIBUTING.md` asks for that agreement on the issue first, because public API is really hard to walk back once it's out. Could you please write a proposal there and tag @PostHog/team-client-libraries, including whether this belongs on `Client` at all, and then land the implementation against whatever we agree on?

.changeset/reusable-client-flush.md
```diff
@@ -0,0 +1,7 @@
+---
+"posthog-go": major
```
marandaneto: this is aditive so just a minor is ok

posthog.go
marandaneto: @ioannisj spec already exists https://github.com/PostHog/posthog-go/issues/229
most SDKs already have a flush method
not sure why we'd need to agree on an API that already exists? mind clarifying?

.changeset/reusable-client-flush.md
marandaneto: Changed the changeset from major to minor in 2b70b48c208a2124073c091805ff3d1b284674c0. Verified the changeset frontmatter and `git diff --check`; the only content change is the release type. All three PR commits now have verified signatures.

marandaneto (approved):

marandaneto:
thanks @cristiangirlea 
@turnipdabeets will get this released asap

cristiangirlea:
@marandaneto thank you also, looking forward for more contributions

